PR #29, opened 2017-11-06
Add varibales to count volume registration in prometheus

utkarshmani1997:
1. Why is this change necessary ?
- To expose size at `/v1/stats` endpoint
- To monitor `/v1/register`

2. How does this change address the issue ?
- Adds a method bind with controller to get size
- Adds prometheus counter variables that is used to count the
  no of registration so far.

3. How to verify this change ?
- Run `curl controllerIP:Port/v1/stats` while running controller
  locally or with openebs-operator.

4. What side effects does this change have ?
- none

5. Other details
- fix: [#858](https://github.com/openebs/openebs/issues/858)
  fix: [#266](https://github.com/openebs/openebs/issues/266)

### DIFF
--- a/controller/rest/model.go
+++ b/controller/rest/model.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/openebs/jiva/controller"
 	"github.com/openebs/jiva/types"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/rancher/go-rancher/api"
 	"github.com/rancher/go-rancher/client"
 )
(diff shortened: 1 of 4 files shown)


kmova:
@utkarshmani1997  .. why is this PR showing changed files in vendor/..gotgt ? -- did you modify those files?

kmova:
@utkarshmani1997  .. why is this PR showing changed files in vendor/..gotgt ? -- did you modify those files?

prateekpandey14:
Looks like he added updated vendor files (gotgt), which should not be the part of commit.

controller/rest/model.go
```diff
@@ -63,6 +64,7 @@ type VolumeStats struct {
 	UsedLogicalBlocks string `json:"UsedLogicalBlocks"`
 	UsedBlocks        string `json:"UsedBlocks"`
 	SectorSize        string `json:"SectorSize"`
+	Size              string `json:"Size"`
```
payes: Have we updated the corresponding structure in maya ??

utkarshmani1997: yes

PR #42, opened 2018-03-20
Enhance jiva for clone feature 

prateekpandey14:
Enhance jiva for clone feature  …
1. This PR will add sync and rebuild capabilities in jiva.
2. Now clone replica can be created from the running replicas and can be further use as RO/RW mode.
3. There is 2 new flags has been added, which has to passed while starting a clone replica with a new jiva-controller.
- `cloneIP` : This is frontendIP of running controller, which will be used to make a clone request will be passed as a argument while starting clone replica.
- `type`: Type of replica, now we can pass type as `clone` to trigger the clone API of jiva, which says the new replica is getting added as clone not as usual replica.

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openebs/jiva/replica"
 	"github.com/openebs/jiva/replica/rest"
 	"github.com/openebs/jiva/replica/rpc"
+	"github.com/openebs/jiva/sync"
 	"github.com/openebs/jiva/util"
 )
 
@@ -32,6 +33,14 @@ func ReplicaCmd() cli.Command {
 			},
 			cli.StringFlag{
 				Name:  "frontendIP",
+				Value: "",
+			},
+			cli.StringFlag{
+				Name:  "cloneIP",
+				Value: "",
+			},
+			cli.StringFlag{
+				Name:  "snapName",
 				Value: "",
 			},
 			cli.StringFlag{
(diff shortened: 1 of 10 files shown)


sync/sync.go
```diff
@@ -207,6 +207,83 @@ Register:
...
+			return err
+		}
+	resetRebuildingTrue:
+		logrus.Infof("Set Rebuild state: true")
+
+		if err := toClient.SetRebuilding(true); err != nil {
+			time.Sleep(2 * time.Second)
+			goto resetRebuildingTrue
+		}
+		for i, name := range chain {
```
payes: What about meta files, there is no check for that.

sync/sync.go
```diff
@@ -207,6 +207,83 @@ Register:
...
+			if name == snapName+".img" {
```
payes: What about meta files, there is no check for that.

app/replica.go
```diff
@@ -190,6 +212,16 @@ func startReplica(c *cli.Context) error {
...
 		go AutoConfigureReplica(s, frontendIP, "tcp://"+address, replicaType)
 	}
+	if replicaType == "clone" {
+		logrus.Infof("Starting clone process using frontend %v", cloneIP)
+		if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+			return err
+		}
+	}
+	for s.Replica() == nil {
+		time.Sleep(2 * time.Second)
```
payes: Why was this sleep required.

PR #44, opened 2018-03-27
Jiva replica clone feature

prateekpandey14:
Enhance jiva for clone feature  …
1. This PR will add sync and rebuild capabilities in jiva as clone feature.
2. Now clone volume can be created from the running replicas volume and can be further used as RO/RW mode.
3. There is 3 new flags has been added, which has to passed while starting a clone replica with a new jiva-controller, to be clone to a particular snapshot.
- `cloneIP` : This is frontendIP of running controller, which will be used to make a clone request will be passed as a argument while starting clone replica.
- `clone`: Type of replica, now we can pass type as `clone` to trigger the clone API of jiva, which says the new replica is getting added as clone not as usual replica.
- `snapName` : Name of the  snapshot which will be synced to a new volume.

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -18,6 +18,7 @@ import (
 	"github.com/openebs/jiva/replica"
 	"github.com/openebs/jiva/replica/rest"
 	"github.com/openebs/jiva/replica/rpc"
+	"github.com/openebs/jiva/sync"
 	"github.com/openebs/jiva/util"
 )
 
@@ -32,6 +33,14 @@ func ReplicaCmd() cli.Command {
 			},
 			cli.StringFlag{
 				Name:  "frontendIP",
+				Value: "",
+			},
+			cli.StringFlag{
+				Name:  "cloneIP",
+				Value: "",
+			},
+			cli.StringFlag{
+				Name:  "snapName",
 				Value: "",
 			},
 			cli.StringFlag{
@@ -92,6 +101,19 @@ checkagain:
 	}
 }
 
+func CloneReplica(s *replica.Server, address string, cloneIP string, snapName string) error {
+	var err error
+	url := "http://" + cloneIP + ":9501"
+	task := sync.NewTask(url)
+	if err = task.CloneReplica(url, address, cloneIP, snapName); err != nil {
+		return err
+	}
+	if s.Replica() != nil {
+		s.Replica().SetCloneStatus("completed")
+	}
+	return err
+}
+
 func startReplica(c *cli.Context) error {
 	if c.NArg() != 1 {
 		return errors.New("directory name is required")
@@ -108,6 +130,8 @@ func startReplica(c *cli.Context) error {
 
 	address := c.String("listen")
 	frontendIP := c.String("frontendIP")
+	cloneIP := c.String("cloneIP")
+	snapName := c.String("snapName")
 	size := c.String("size")
 	if size != "" {
 		//Units bails with an error size is provided with i, like Gi
@@ -189,6 +213,22 @@ func startReplica(c *cli.Context) error {
 			address = "localhost:9502"
 		}
 		go AutoConfigureReplica(s, frontendIP, "tcp://"+address, replicaType)
+	}
+	for s.Replica() == nil {
+		time.Sleep(2 * time.Second)
+	}
+	if replicaType == "clone" && snapName != "" {
+		logrus.Infof("Starting clone process\n")
+		status := s.Replica().GetCloneStatus()
+		if status != "completed" {
+			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				return err
+			}
+		}
+		s.Replica().SetCloneStatus("completed")
+		logrus.Infof("Clone process completed successfully\n")
+	} else {
+		s.Replica().SetCloneStatus("NA")
 	}
 
 	return <-resp
--- a/app/replica.go
+++ b/app/replica.go
@@ -221,7 +221,9 @@ func startReplica(c *cli.Context) error {
 		logrus.Infof("Starting clone process\n")
 		status := s.Replica().GetCloneStatus()
 		if status != "completed" {
+			s.Replica().SetCloneStatus("inProgress")
 			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				s.Replica().SetCloneStatus("error")
 				return err
 			}
 		}
--- a/backend/file/file.go
+++ b/backend/file/file.go
@@ -86,6 +86,10 @@ func (f *Wrapper) GetMonitorChannel() types.MonitorChannel {
 	return nil
 }
 
+func (f *Wrapper) GetCloneStatus() (string, error) {
+	return "", nil
+}
+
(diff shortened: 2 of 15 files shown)


replica/replica.go
```diff
@@ -856,11 +869,13 @@ func (r *Replica) openLiveChain() error {
 	}
 
 	for i := len(chain) - 1; i >= 0; i-- {
+
```
gila: this change is not needed?

replica/replica.go
```diff
@@ -856,11 +869,13 @@ func (r *Replica) openLiveChain() error {
...
 		parent := chain[i]
 		f, err := r.openFile(parent, 0)
 		if err != nil {
 			return err
 		}
+
```
gila: same

replica/replica.go
```diff
@@ -900,6 +915,7 @@ func (r *Replica) readMetadata() (bool, error) {
 			}
 		}
 	}
+
```
gila: same

app/replica.go
```diff
@@ -190,6 +214,22 @@ func startReplica(c *cli.Context) error {
...
+	}
+	if replicaType == "clone" && snapName != "" {
+		logrus.Infof("Starting clone process\n")
+		status := s.Replica().GetCloneStatus()
+		if status != "completed" {
+			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				return err
+			}
+		}
+		s.Replica().SetCloneStatus("completed")
```
gila: What type of statuses do we have?  just completed and NA? what about in progress? error? etc?

backend/file/file.go
```diff
@@ -86,6 +86,10 @@ func (f *Wrapper) GetMonitorChannel() types.MonitorChannel {
 	return nil
 }
 
+func (f *Wrapper) GetCloneStatus() (string, error) {
```
gila: Why is error a string and not a number/enum?

controller/control.go
```diff
@@ -583,6 +583,12 @@ func (c *Controller) Start(addresses ...string) error {
 		if err := c.addReplicaNoLock(newBackend, address, false); err != nil {
 			return err
 		}
+	getCloneStatus:
+		if status, _ := c.backend.GetCloneStatus(address); status != "completed" && status != "NA" {
```
gila: status must have 2 values at the same time?

replica/replica.go
prateekpandey14: Not needed , i will remove that space.

replica/replica.go
prateekpandey14: ok

controller/control.go
prateekpandey14: Here we are checking that current status of clone, it can be either `completed` or `NA`, if either one of them is not there that means sync is going on, then check it again in 2 sec interval

replica/replica.go
prateekpandey14: will remove, may be added by the editor

backend/file/file.go
prateekpandey14: error is a built in type, but custom error can be returned, that will contains the actual error

prateekpandey14:
@gila There are no unit test cases in jiva for anything, we have to write all of them but can be done in upcoming PR's

controller/control.go
gila: thats not what the code says... or means && something else in go?

app/replica.go
gila: So why not add state that's in progress? If someone else needs to find out he/she would need to find this PR to understand the reasoning.

backend/file/file.go
gila: Ok..

controller/control.go
payes: If the replica type is clone, then we wait over here for its status to be turned to "completed".
If the replica is an original replica (not a cloned version), we wait for the status to be set to "NA(Not Applicable).

gila: status != "completed" && status != "NA" this reads to me that the status should be NOT completed and NOT NA, so what state does this imply? this is a state that does not exist right now?

ci/start_init_test.sh
```diff
@@ -74,16 +80,81 @@ if [ "$x"!="" ]; then
...
+sudo docker run -d -it --net stg-net --ip 172.18.0.6 -P --expose 9502-9504 -v /tmp/vol3:/vol3 $JI launch replica --type clone --snapName snap1 --cloneIP 172.18.0.2 --frontendIP 172.18.0.5 --listen 172.18.0.6:9502 --size 2g /vol3
+
+clonestatus=`curl http://172.18.0.6:9502/v1/replicas/1 | jq '.clonestatus' | tr -d '"'`
+
+i=0
+while [ -z $clonestatus ]; do
+        sleep 3
+	clonestatus=`curl http://172.18.0.6:9502/v1/replicas/1 | jq '.clonestatus' | tr -d '"'`
+        i=`expr $i + 1`
+        if [ $i -eq 20 ]; then
```
gila: Why not run until we get an error, after all, we added error cases yes?

ci/start_init_test.sh
```diff
@@ -74,16 +80,81 @@ if [ "$x"!="" ]; then
...
 
+	# TEST#1: Perform simple data-integrity check on Jiva Vol
+	hash3=$(sudo md5sum /mnt/store/file1 | awk '{print $1}')
+	if [ $hash1 == $hash3 ]; then echo "DI Test: PASSED"
+	else
+		echo "DI Test: FAILED"; exit 1
+	fi
+else
+	echo "Unable to detect iSCSI device, login failed"; exit 1
+fi
```
gila: How is this test run automatically?

app/replica.go
```diff
@@ -221,7 +221,9 @@ func startReplica(c *cli.Context) error {
 		logrus.Infof("Starting clone process\n")
 		status := s.Replica().GetCloneStatus()
 		if status != "completed" {
+			s.Replica().SetCloneStatus("inProgress")
 			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				s.Replica().SetCloneStatus("error")
```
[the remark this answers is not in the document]
payes: Whenever there is an error we print it on the terminal, which is captured by kubectl logs.
Over here cloneStatus is the state which is being stored in the metadata file. Would it be proper to write the reason for the state also in the metadata file.

ci/start_init_test.sh
payes: We are exiting the container if the replica clone fails. Therefore we will and be able to get the status of the replica and this will be stuck in infinite loop. Or we could also check for the http error.

ci/start_init_test.sh
payes: I didn't get the question. If you are asking from where this shell script is triggered, its from the Makefile, after the image is built.

controller/control.go
payes: It means the state has not been set till now by the replica. i.e the status returned is "".
Or, inprogress and error states have also been added.
A new commit has been added where inprogress/""/error states are being checked instead of above logic to be more clearer.

PR #45, opened 2018-03-28
Fixes for replica failures

payes:
This pull request fixes the following issue:
The controller was not getting converted back to read-write mode when the replica came back up in case of single replica. The controller had to be restarted as a work around.

### DIFF
--- a/controller/rest/model.go
+++ b/controller/rest/model.go
@@ -21,6 +21,7 @@ type Volume struct {
 	client.Resource
 	Name         string `json:"name"`
 	ReplicaCount int    `json:"replicaCount"`
+	ReadOnly     string `json:"readOnly"`
 }
 
 type VolumeCollection struct {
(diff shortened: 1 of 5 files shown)


controller/rest/model.go
```diff
@@ -21,6 +21,7 @@ type Volume struct {
 	client.Resource
 	Name         string `json:"name"`
 	ReplicaCount int    `json:"replicaCount"`
+	ReadOnly     string `json:"readOnly"`
```
gila: why not a boolean? JSON supports booleans just fine.

controller/rest/model.go
```diff
@@ -105,7 +106,13 @@ type RegReplica struct {
...
 }
 
-func NewVolume(context *api.ApiContext, name string, replicas int) *Volume {
+func NewVolume(context *api.ApiContext, name string, readOnly bool, replicas int) *Volume {
+	var ReadOnly string
+	if readOnly {
+		ReadOnly = "true"
+	} else {
+		ReadOnly = "false"
+	}
```
gila: See my previous comment about bools :)

ksatchit:
Fixes https://github.com/openebs/openebs/issues/1383

PR #54, opened 2018-05-30
Add test case for 3 replicas and reduce R/W timeout and retry values

payes:

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -15,9 +15,9 @@ var (
 	ErrRWTimeout   = errors.New("r/w timeout")
 	ErrPingTimeout = errors.New("Ping timeout")
 
-	opRetries       = 1
-	opReadTimeout   = 5 * time.Second // client read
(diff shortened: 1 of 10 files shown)


utkarshmani1997:
LGTM

rpc/client.go
```diff
@@ -15,9 +15,9 @@ var (
 	ErrRWTimeout   = errors.New("r/w timeout")
 	ErrPingTimeout = errors.New("Ping timeout")
 
-	opRetries       = 4
-	opReadTimeout   = 15 * time.Second // client read
-	opWriteTimeout  = 15 * time.Second // client write
+	opRetries       = 1
```
kmova: @payes Can you add comment on how these fields are used, why we have to change these time out values?

payes: These values were based on the scsi timeout values which were seen earlier from the initiator side while testing istgt. But these changes will be reverted back, others are also not happy with these timeout related changes.

PR #55, opened 2018-06-07
 Fix quorum related issues and a race condition

payes:
1)Reply with success when IOs are written to atleast 51% of replicas.
  Add delays in through signals in read/write code to simulate the above case.
2)Attempt replica close before trying to connect to controller.
  Replica can only connect to controller in closed state.
3)Add delays through signals in replica addition path to simulate race condition.

### DIFF
--- a/app/controller.go
+++ b/app/controller.go
@@ -93,18 +93,18 @@ func startController(c *cli.Context) error {
 		}
 	}
 
-	signal_chan := make(chan os.Signal, 5)
-	signal.Notify(signal_chan, syscall.SIGUSR1, syscall.SIGUSR2)
+	signalChan := make(chan os.Signal, 5)
+	signal.Notify(signalChan, syscall.SIGUSR1, syscall.SIGUSR2)
 	go func() {
 		for {
-			s := <-signal_chan
+			s := <-signalChan
 			switch s {
 			case syscall.SIGUSR1:
-				controller.DelayInSec += 2
+				controller.Delay += 2
 			case syscall.SIGUSR2:
-				controller.DelayInSec -= 2
+				controller.Delay -= 2
 			}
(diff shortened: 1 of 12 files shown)


app/controller.go
```diff
@@ -91,6 +93,21 @@ func startController(c *cli.Context) error {
...
+	signal_chan := make(chan os.Signal, 5)
+	signal.Notify(signal_chan, syscall.SIGUSR1, syscall.SIGUSR2)
+	go func() {
+		for {
+			s := <-signal_chan
+			switch s {
+			case syscall.SIGUSR1:
+				controller.DelayInSec += 2
+			case syscall.SIGUSR2:
+				controller.DelayInSec += 2
```
pawanpraka1: controller.DelayInSec -= 2?

payes: thanks

controller/multi_writer_at.go
```diff
@@ -80,9 +82,27 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
...
+			if err1 != nil {
+				replicaErrCount++
+			}
+		}
+		for _, err1 := range quorumErrs {
+			if err1 != nil {
+				quorumErrCount++
+			}
+		}
+		if (len(m.writers)-replicaErrCount >= len(m.writers)/2) &&
```
pawanpraka1: should it be > instead of >= as in case of 4 replica we want IO to be successful on n/2 + 1 replica i.e. 3?

payes: Resolved in latest commit and ci test case is also being rechecked to figure out why this issue was not caught.

payes: There was no test case for this condition to check for failure.

PR #57, opened 2018-06-17
Replica/Controller crash related fixes

payes:
Avoid sending signal to monitorchan until replica is connected to controller
Add read locks to protect replica data structure


pawanpraka1:
LGTM

PR #66, opened 2018-06-29
Remove retries, and fix a race condition in logging

payes:
Controller does retrying IO to replica when IO gets timed out on replica. Response received from replica for IOs after timeout are not considered as success response. Also, controller doesn't take new IOs until response (either error/timeout/success) is received from all the replicas. This will timeout the client.
This PR is to make timeout retry count as 0, so that, retries won't happen on timed out IOs.

A routine will be waiting for an IO response from client, and also, when it times out, it prints the error message by taking the length from the message. At the same time, if error happens on that IO, another routine dedicated for client can modify the message. This leads to race between these two go routines.
This PR covers the race by not printing the length of the message.

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -14,8 +14,8 @@ var (
 	//ErrRWTimeout r/w operation timeout
 	ErrRWTimeout   = errors.New("r/w timeout")
 	ErrPingTimeout = errors.New("Ping timeout")
-
-	opRetries       = 4
+	//Retries are not required as the reader on msg->complete has already exited
+	opRetries       = 0
 	opReadTimeout   = 15 * time.Second // client read
 	opWriteTimeout  = 15 * time.Second // client write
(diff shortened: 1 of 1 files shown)


rpc/client.go
```diff
@@ -142,22 +142,22 @@ func (c *Client) operation(op uint32, buf []byte, offset int64) (int, error) {
 		case <-timeout:
 			switch msg.Type {
 			case TypeRead:
-				logrus.Errorln("Read timeout on replica", c.TargetID(), "seq=", msg.Seq, "size=", len(msg.Data)/1024, "(kB)")
+				logrus.Errorln("Read timeout on replica ", c.TargetID(), c.peerAddr, "seq=", msg.Seq)
```
vishnuitta: We can also print length of msg.Data by having a local variable.

PR #68, opened 2018-06-29
rpc: Eliminate empty data transfer between client/server

payes:
We don't need data buffer for read request or write response, because they're
empty. For this purpose, we introduced `Size` in message struct. It will be the
intended size of read or write.

This PR has been picked from upstream.
https://github.com/rancher/longhorn-engine/commit/feb9f4b63e22702318a1bedd0912939fa9661e4a

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -110,7 +110,11 @@ func (c *Client) operation(op uint32, buf []byte, offset int64) (int, error) {
 			Complete: make(chan struct{}, 1),
 			Type:     op,
 			Offset:   offset,
-			Data:     buf,
+			Size:     uint32(len(buf)),
+			Data:     nil,
+		}
+		if op == TypeWrite {
+			msg.Data = buf
 		}
 
 		timeout := func(op uint32) <-chan time.Time {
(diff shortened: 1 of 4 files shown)


rpc/client.go
```diff
@@ -132,15 +136,19 @@ func (c *Client) operation(op uint32, buf []byte, offset int64) (int, error) {
 
 		select {
 		case <-msg.Complete:
+			// Only copy the message if a read is requested
+			if op == TypeRead && (msg.Type == TypeResponse || msg.Type == TypeEOF) {
+				copy(buf, msg.Data)
+			}
 			if msg.Type == TypeError {
 				return 0, errors.New(string(msg.Data))
```
pawanpraka1: why are we returning msg.Data as error?

rpc/server.go
```diff
@@ -115,13 +116,19 @@ func (s *Server) handleUpdate(msg *Message) {
...
 	msg.MagicVersion = MagicVersion
+	msg.Size = uint32(len(msg.Data))
+	if msg.Type == TypeWrite {
+		msg.Data = nil
+	}
 	msg.Type = TypeResponse
 	if err == io.EOF {
-		msg.Data = msg.Data[:count]
 		msg.Type = TypeEOF
+		msg.Data = msg.Data[:count]
```
pawanpraka1: in case msg.Type == TypeWrite , msg.Data will be nil. can we do msg.Data[:count] in this case? Probably we should put a chekck if it's not TypeWrite then only set msg.data.

rpc/client.go
pawanpraka1: ok, we are returning err only in case of TypeError, so it' looks fine.

PR #71, opened 2018-07-04
Fix race conditions in Read/Write calls

payes:


controller/control.go
```diff
@@ -475,6 +475,52 @@ func (c *Controller) RemoveReplica(address string) error {
 	return nil
 }
 
+func (c *Controller) RemoveReplicaNoLock(address string) error {
```
vishnuitta: avoid duplicate code

PR #75, opened 2018-07-05
Fix quorum related issues and a race condition

utkarshmani1997:
This commit is cherry-picked from https://github.com/openebs/jiva/commit/59b33519b68a85ee573a094d1056367d4bc87130
* Fix quorum related issues and a race condition
1)Reply with success when IOs are written to atleast 51% of replicas.
  Add delays in through signals in read/write code to simulate the above case.
2)Attempt replica close before trying to connect to controller.
  Replica can only connect to controller in closed state.
3)Add delays through signals in replica addition path to simulate race condition.
* Add test case for 3 replicas and reduce R/W timeout and retry values

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ _run_ci:
 	@echo ""
 	@echo "INFO:\t..... run ci over jiva image"
 	@echo ""
-	./ci/start_init_test.sh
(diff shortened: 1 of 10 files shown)


utkarshmani1997:
This cherry-pick is not required, so closing the PR

PR #80, opened 2018-07-10
Rectify string comparision while removing replica from registered list and add logs

payes:
There was a problem while clearing replica details from controller when replica goes down.
The address passed in RemoveReplica function contains the strings "tcp://" and ":9502".

But the registered replica list contains only ip addresses and we were checking for this ip substring in the passed address which is tcp://<ip str>:<port>.
So in case of two replica having IPs 10.10.10.10 and 10.10.10.1, the substring comparison of 10.10.10.1 will also be true for the replica with address tcp://10.10.10.10:9502 and we will remove wrong (or both) the replica in this case.

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -427,7 +427,12 @@ func (c *Controller) hasReplica(address string) bool {
 }
 
 func (c *Controller) RemoveReplicaNoLock(address string) error {
+	var foundregrep int
+	var prev bool
+
+	logrus.Infof("RemoveReplica %v ReplicasAdded:%v FrontendState:%v", address, len(c.replicas), c.frontend.State())
 	if !c.hasReplica(address) {
+		logrus.Infof("RemoveReplica %v not found", address)
 		return nil
 	}
 	for i, r := range c.replicas {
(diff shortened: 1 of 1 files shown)


controller/control.go
```diff
@@ -443,32 +448,49 @@ func (c *Controller) RemoveReplicaNoLock(address string) error {
...
+				logrus.Infof("RemoveReplica ToRemove: %v Found: %v", address, regrep)
+				if address == "tcp://"+regrep+":9502" {
 					delete(c.RegisteredReplicas, regrep)
+					foundregrep = 1
+					break
 				}
 			}
+			if foundregrep == 0 {
+				logrus.Infof("RemoveReplica %v not found in registered replicas", address)
+			}
```
vishnuitta: should break from here

payes: Break is not required as all replicas will not go through the registration process , hence they will not be present in registered replica list. If the volume is already running, then atleast one of the replicas will be in R/W state. Therefore, the registration process to compare the quorum can be avoided.

PR #83, opened 2018-07-11
Fetch Replication factor from env and make volume R/W only when >50% replicas are in R/W mode

payes:
If replication factor is received from control plane, data corruption can be avoided.
Scenario:
2 Replicas are present.
1) Replica_1 is connected to controller and IOs have been started, but replica 2 did not connect till now.
2) Controller restarts and replica 2 connects before replica 1. Revision counter and peerCount at Replica_2 are 0, so controller thinks that it is a new deployment and starts the IOs on this replica.
3) Replica_1 now connects back to controller and its data is synced, assuming Replica_1 as source.
Now the data which was written earlier at Replica_1 is lost.  

Other fixes:
- There is unnecessary creation of TempServer and TempReplica during repeated registration. Removed the unnecessary creation.

### DIFF
--- a/app/controller.go
+++ b/app/controller.go
@@ -6,6 +6,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"strconv"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
@@ -66,6 +67,12 @@ func startController(c *cli.Context) error {
 		return errors.New("volume name is required")
 	}
 	name := c.Args()[0]
+	replicationFactor, _ := strconv.ParseInt(os.Getenv("REPLICATION_FACTOR"), 10, 32)
+	if replicationFactor == 0 {
+		logrus.Infof("REPLICATION_FACTOR env not set")
+	} else {
+		logrus.Infof("REPLICATION_FACTOR: %v", replicationFactor)
+	}
 
 	if !util.ValidVolumeName(name) {
 		return errors.New("invalid target name")
@@ -100,7 +107,7 @@ func startController(c *cli.Context) error {
 		frontend = f
 	}
 
-	control := controller.NewController(name, frontendIP, clusterIP, dynamic.New(factories), frontend)
+	control := controller.NewController(name, frontendIP, clusterIP, dynamic.New(factories), frontend, int(replicationFactor))
 	server := rest.NewServer(control)
 	router := http.Handler(rest.NewRouter(server))
 
--- a/app/replica.go
+++ b/app/replica.go
@@ -85,7 +85,9 @@ func CheckReplicaState(frontendIP string, replicaIP string) (string, error) {
 
 func AutoConfigureReplica(s *replica.Server, frontendIP string, address string, replicaType string) {
 checkagain:
+	logrus.Infof("Get replica state")
 	state, err := CheckReplicaState(frontendIP, address)
+	logrus.Infof("checkreplicastate %v %v", state, err)
 	if err == nil && (state == "" || state == "ERR") {
 		s.Close(false)
 	} else {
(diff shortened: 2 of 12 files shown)


controller/control.go
```diff
@@ -242,13 +253,8 @@ func (c *Controller) registerReplica(register types.RegReplica) error {
...
 	}
 
-	if ((len(c.RegisteredReplicas)) >= c.replicaCount/2) && ((len(c.RegisteredReplicas) + len(c.RegisteredQuorumReplicas)) > (c.quorumReplicaCount+c.replicaCount)/2) {
-		c.signalToAdd()
-		c.StartSignalled = true
-		return nil
-	}
-
-	if c.RegisteredReplicas[c.MaxRevReplica].PeerDetail.ReplicaCount <= 1 {
+	if (len(c.RegisteredReplicas) > c.replicaCount/2) &&
```
vishnuitta: we need to check for 'max' here also

app/controller.go
```diff
@@ -66,6 +67,12 @@ func startController(c *cli.Context) error {
 		return errors.New("volume name is required")
 	}
 	name := c.Args()[0]
+	replicationFactor, _ := strconv.ParseInt(os.Getenv("REPLICATION_FACTOR"), 10, 32)
```
vishnuitta: this code change should be enough if we set c.replicaCount to replicationFactor here itself. No need of other changes given in this PR.

app/replica.go
```diff
@@ -85,7 +85,9 @@ func CheckReplicaState(frontendIP string, replicaIP string) (string, error) {
 
 func AutoConfigureReplica(s *replica.Server, frontendIP string, address string, replicaType string) {
 checkagain:
+	logrus.Infof("Get replica state")
```
vishnuitta: this log can be removed

ci/start_init_test.sh
```diff
@@ -186,16 +189,15 @@ verify_controller_rep_state() {
...
 	done
 	echo $3 " -- failed"
 	collect_logs_and_exit
 }
 
 # start_controller CONTROLLER_IP
 start_controller() {
-	controller_id=$(docker run -d --net stg-net --ip "$1" -P --expose 3260 --expose 9501 --expose 9502-9504 $JI \
-		launch controller --frontend gotgt --frontendIP "$1" "$2")
+	controller_id=$(docker run -d --net stg-net --ip $1 -P --expose 3260 --expose 9501 --expose 9502-9504 $JI env REPLICATION_FACTOR="$3" launch controller --frontend gotgt --frontendIP "$1" "$2")
```
vishnuitta: is there any problem in having the '\', i.e., splitting this big single line to multiple small ones ?

controller/control.go
```diff
@@ -49,11 +62,32 @@ func NewController(name string, frontendIP string, clusterIP string, factory typ
...
 
+func (c *Controller) UpdateVolStatus() {
+	prev := c.ReadOnly
+	var rwReplicaCount int
+	for _, replica := range c.replicas {
+		if replica.Mode == "RW" {
+			rwReplicaCount++
+		}
+	}
+	if (rwReplicaCount > (c.replicaCount+c.quorumReplicaCount)/2) &&
```
vishnuitta: code would be readable if you have like:
```if (rwReplicaCount >= ((c.replicaCount + c.quorumReplicaCount)/2 + 1) {
        if (c.ReadOnly == true) {
            logrus....
            c.ReadOnly = false;
        }
     } else {
        if (c.ReadOnly == true) {
            logrus....
            c.ReadOnly = false;
        }
    }
```

controller/control.go
```diff
@@ -203,8 +230,12 @@ func (c *Controller) registerReplica(register types.RegReplica) error {
 	if !ok {
 		_, ok = c.RegisteredQuorumReplicas[register.Address]
 		if ok {
+			logrus.Infof("Quorum replica Address %v already present in registered list", register.Address)
 			return nil
 		}
+	} else {
+		//TODO Check if it is better to return from here
```
vishnuitta: remove this TODO

PR #84, opened 2018-07-12
[TA1772] Testcase for last replica being in WO mode

vishnuitta:
This PR is to simulate the case where the last replica that is connected to controller is in WO mode.
In this scenario, replica (which is in WO mode) closes itself when sync fails and re-registers with controller.
New code in `RemoveReplicaNoLock` that is handling this case specially is not required. This PR reverts that code.
Few other changes to 'ci' test:
- Increased the wait time to 200 seconds or 250 seconds in reaching the 'pass' state
- Collecting stack traces, replicas/volumes information during failure case
- Taking the head of the docker image during docker creation

### DIFF
--- a/ci/start_init_test.sh
+++ b/ci/start_init_test.sh
@@ -12,21 +12,26 @@ collect_logs_and_exit() {
 	echo "--------------------------docker ps -a-------------------------------------"
 	docker ps -a
 
+	echo "--------------------------CONTROLLER REST output---------------------------"
+	curl http://$CONTROLLER_IP:9501/v1/volumes | jq
+	curl http://$CONTROLLER_IP:9501/v1/replicas | jq
+	echo "--------------------------REPLICA 1 LOGS ----------------------------------"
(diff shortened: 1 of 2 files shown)


ci/start_init_test.sh
```diff
@@ -87,60 +220,70 @@ get_replica_count() {
...
+	count=0
+	while [ "$count" != 10 ]; do
+		docker stop $replica1_id
+
+		docker start $replica1_id &
+		sleep `echo "$count * 0.2" | bc`
+		docker stop $replica2_id
+		# Replica1 might be in Registering mode with status as 'closed' or its rebuild is done with mode as 'RW'
+		verify_rep_state 1 "Replica1 status after restarting it, and stopping other one in 2 replicas case" "$REPLICA_IP1" "RW"
+
```
pawanpraka1: can we check verify_vol_status status as RO here?

pawanpraka1: never mind, I see that test case is added at 279.

PR #88, opened 2018-07-16
[TA1899][TA1900][TA1901] Adding logs in controller during replica reg…

vishnuitta:
…istration/start/add

This PR is to add logs to controller for better debugging of replica registration / replica start, and, replica add process.

Apart from logs, fixed an API 'controller.canAdd' that checks whether given address is already added or not. It is returning NULL which is not supposed to be the case as the returned error is being used by the caller of it to return to its caller.

Another change is - made fatalError if a replica address is found more than once in replica list during setting the mode of replica.

Signed-off-by: Vishnu Itta <vitta@mayadata.io>

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"math"
 	"sync"
 	"time"
 
@@ -388,11 +389,11 @@ func (c *Controller) addQuorumReplicaNoLock(newBackend types.Backend, address st
 	return nil
 }
 
-/*
-* No need to add prints in this function.
-* Make sure caller of this takes care of printing error
- */
 func (c *Controller) addReplicaNoLock(newBackend types.Backend, address string, snapshot bool) error {
+	/*
+	 * No need to add prints in this function.
+	 * Make sure caller of this takes care of printing error
+	 */
 	if ok, err := c.canAdd(address); !ok {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


controller/control.go
```diff
@@ -851,15 +893,16 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
...
 	if monitorChan == nil {
+		logrus.Errorf("cannot monitor %s.. chan is null", address)
 		return
 	}
 
 	logrus.Infof("Start monitoring %v", address)
 	err := <-monitorChan
 	c.Lock()
 	defer c.Unlock()
+	logrus.Errorf("Backend %v monitoring failed, mark as ERR: %v", address, err)
```
pawanpraka1: put this inside below if error check.

controller/multi_writer_at.go
```diff
@@ -53,6 +53,8 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
 	var errors MultiWriterError
 	var multiWriterMtx sync.Mutex
 
+	multiWriterMtx = sync.Mutex{}
```
pawanpraka1: not needed.

controller/control.go
vishnuitta: done

controller/multi_writer_at.go
vishnuitta: done

PR #93, opened 2018-07-19
[TA328] add error logs if clone is not completed.

utkarshmani1997:
On branch fix-clone-logs
Changes to be committed:
	modified:   ../../../app/replica.go

-  Error was ignored while setting the status of cloned replcas.

-  Added the logs and err var to catch those errors.

-  None

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"errors"
-	"fmt"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


app/replica.go
```diff
@@ -233,19 +234,25 @@ func startReplica(c *cli.Context) error {
...
-			s.Replica().SetCloneStatus("inProgress")
-			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
-				logrus.Infof("Set clone status as error")
-				s.Replica().SetCloneStatus("error")
+			if err := s.Replica().SetCloneStatus("inProgress"); err != nil {
 				return err
 			}
+			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				logrus.Error("Error in cloning replica, found error ", err.Error())
+				logrus.Info("Setting the status of clone as error")
```
pawanpraka1: why we need this after having Error log?

app/replica.go
```diff
@@ -233,19 +234,25 @@ func startReplica(c *cli.Context) error {
...
+				return s.Replica().SetCloneStatus("error")
+			}
 		}
 		logrus.Infof("Set clone status as Completed")
-		s.Replica().SetCloneStatus("completed")
+		if err := s.Replica().SetCloneStatus("completed"); err != nil {
+			return fmt.Errorf("Error in setting the status of clone as complete, found error %s", err.Error())
```
pawanpraka1: return err instead of creating a new error.

app/replica.go
utkarshmani1997: We are adding this info to notify that we are setting the clone status as error. Should we remove this ?

app/replica.go
utkarshmani1997: Because we don't know where this error is happening due to hierarchy of calls, which includes some system calls also.

app/replica.go
```diff
@@ -233,19 +234,24 @@ func startReplica(c *cli.Context) error {
...
-			s.Replica().SetCloneStatus("inProgress")
-			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
-				logrus.Infof("Set clone status as error")
-				s.Replica().SetCloneStatus("error")
+			if err := s.Replica().SetCloneStatus("inProgress"); err != nil {
 				return err
 			}
+			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
+				logrus.Error("Error in cloning replica, found error ", err.Error())
+				return s.Replica().SetCloneStatus("error")
```
vishnuitta: this overwrites the error that got for CloneReplica() API. Needs a fix here.

app/replica.go
```diff
@@ -233,19 +234,24 @@ func startReplica(c *cli.Context) error {
 		status := s.Replica().GetCloneStatus()
 		if status != "completed" {
 			logrus.Infof("Set clone status as inProgress")
-			s.Replica().SetCloneStatus("inProgress")
-			if err = CloneReplica(s, "tcp://"+address, cloneIP, snapName); err != nil {
-				logrus.Infof("Set clone status as error")
-				s.Replica().SetCloneStatus("error")
+			if err := s.Replica().SetCloneStatus("inProgress"); err != nil {
```
vishnuitta: why do we need ':=' here?

app/replica.go
utkarshmani1997: fixed

app/replica.go
utkarshmani1997: fixed

PR #97, opened 2018-07-26
Logs in REST API handling in controller

payes:
This PR adds logs to REST API handling in controller.


controller/control.go
```diff
@@ -523,10 +519,14 @@ func (c *Controller) RemoveReplica(address string) error {
 }
 
 func (c *Controller) ListReplicas() []types.Replica {
+	c.Lock()
```
vishnuitta: this will not provide any extra lock protection, if I'm not wrong

PR #99, opened 2018-07-26
Lock protection for replica data structures during REST API handling

vishnuitta:
In replicas of jiva volumes, during REST API handling, replica related data structures are used without any lock protection.
This leads to data race between reader and writer of those structures.

This PR is to lock protect of those structures during REST API handling.

This PR also adds logs to track REST APIs and its errors.

### DIFF
--- a/replica/server.go
+++ b/replica/server.go
@@ -187,10 +187,8 @@ func (s *Server) PrevStatus() (State, Info) {
 	return Closed, info
 }
 
-/*
-* Stats() returns the revisionCache and Peerdetails
-* TODO: What to return in Stats and GetUsage if s.r is nil?
- */
+// Stats returns the revisionCache and Peerdetails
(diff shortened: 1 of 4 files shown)


replica/server.go
```diff
@@ -188,14 +189,32 @@ func (s *Server) PrevStatus() (State, Info) {
...
 	}
+	return stats1
 }
 
 func (s *Server) GetUsage() (*types.VolUsage, error) {
-	return s.r.GetUsage()
+	if s.r != nil {
+		return s.r.GetUsage()
+	}
+	return &types.VolUsage{
```
pawanpraka1: why this change? shouldn't we return error in this case?

vishnuitta: error seems to be ignored in the code, which is why just returning error may not be sufficient. It needs nil handling in the callers.
Added a TODO for nil handling

PR #103, opened 2018-08-08
[TA2437] Fix goroutine leaks in jiva

utkarshmani1997:
On branch US2634-fix-goroutine-leaks
 Changes to be committed:
	modified:   replica/rest/router.go
	modified:   replica/rpc/server.go
	modified:   replica/server.go
-  To fix the goroutine leaks in jiva.

-  Added an additional check before handling client requests for verification
   of the expected rpc client from controller.
-  Added an additional field `ClientAddr` in the `replica.Server` which can be
   used for verifying the clients and hence avoid goroutine leak.

-  None

4. How to verify this change?
-  Make request to replica by `curl replicaIP:9503` and open the debug/pprof UI
   and you should not see the increase in the goroutines.

5. Any specific message for reviewer ?
-  None

### DIFF
--- a/ci/start_init_test.sh
+++ b/ci/start_init_test.sh
@@ -205,6 +205,29 @@ verify_controller_quorum() {
 		i=`expr $i + 1`
 	done
 }
+# This verifies the goroutine leaks which happens when a request is made to
+# replica_ip:9503.
+verify_go_routine_leak() {
+    i=0
+    date
+    no_of_goroutine=`curl http://$2:9502/debug/pprof/goroutine?debug=1 | grep goroutine | awk '{ print $4}'`
+    passed=0
+    req_cnt=0
+    while [ "$i" != 30 ]; do
+            curl http://$2:9503
+            i=`expr $i + 1`
+            sleep 2
+    done
+    wait
+    new_no_of_goroutine=`curl http://$2:9502/debug/pprof/goroutine?debug=1 | grep goroutine | awk '{ print $4}'`
+    old=`expr $no_of_goroutine + 3`
+    if [ $new_no_of_goroutine -lt $old ]; then
+             echo $1 --passed
+             return
+    fi
+    echo $1 " -- failed"
+    collect_logs_and_exit
+}
 
 verify_vol_status() {
 	i=0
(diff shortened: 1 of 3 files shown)


replica/rpc/server.go
```diff
@@ -51,12 +54,47 @@ func (s *Server) ListenAndServe() error {
 			continue
 		}
 
-		logrus.Infof("New connection from: %v", conn.RemoteAddr())
+		if len(s.s.ClientAddr) == 0 {
+			logrus.Infof("New connection from: %v", conn.RemoteAddr())
+			s.s.ClientAddr = conn.RemoteAddr().String()
```
vishnuitta: let us set clientAddr during handlePing of rpc/server.go, and use that address for comparison here.

ci/start_init_test.sh
```diff
@@ -287,6 +287,29 @@ verify_rep_state() {
 	collect_logs_and_exit
 }
 
+verify_go_routine_leak() {
```
vishnuitta: indentation is required.
Please add comments.

ci/start_init_test.sh
```diff
@@ -382,6 +405,9 @@ test_two_replica_stop_start() {
 
 	docker start $replica1_id
 	verify_replica_cnt "2" "Two replica count test2"
+    verify_go_routine_leak "when there are 2 replicas, and sending curl request on data address to" "$REPLICA_IP1"
```
vishnuitta: checking on one is fine.

replica/rest/router.go
```diff
@@ -36,6 +36,7 @@ func NewRouter(s *Server) *mux.Router {
 	f := HandleError
 
 	router.Methods("GET").Path("/ping").Handler(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
+		s.s.ClientAddr = req.RemoteAddr
```
vishnuitta: this is not required.

rpc/server.go
```diff
@@ -25,8 +26,9 @@ func NewServer(conn net.Conn, data types.DataProcessor) *Server {
 	}
 }
 
-func (s *Server) SetMonitorChannel(monitorChan chan struct{}) {
+func (s *Server) SetServerAttributes(monitorChan chan struct{}, addr *string) {
 	s.monitorChan = monitorChan
+	s.clientAddr = addr
```
vishnuitta: when do you reset it to nil?

utkarshmani1997: @vishnuitta in the handlePing() function. Here i'm just assigning the address.

PR #105, opened 2018-08-24
[US2662] check if underlying filesystem supports extents mapping.

utkarshmani1997:
On branch US2662-verify-file-extents-mapping
 Changes to be committed:

-  To verify whether the underlying filesystem supports extents mapping.

-  In the earlier code we were ignoring the error which verifies if the
   filesystem supports extents mapping. So now we will catch the error
   and error out the replica.

-  None.

4. How to verify this change?
-  ci is already added to verify the changes.

5. Any specific message for reviewer ?
-  No

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -224,7 +224,11 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.insertBackingFile()
 	r.ReplicaType = replicaType
 
-	PreloadLunMap(&r.volume)
+	if err := PreloadLunMap(&r.volume); err != nil {
+		logrus.Error("underlying file system does not support extent mapping")
+		return r, err
+	}
+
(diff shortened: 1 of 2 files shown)


replica/replica.go
```diff
@@ -224,7 +224,11 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.insertBackingFile()
 	r.ReplicaType = replicaType
 
-	PreloadLunMap(&r.volume)
+	if err := PreloadLunMap(&r.volume); err != nil {
+		logrus.Error("underlying file system does not support extent mapping")
```
payes: Lets also print the actual error received.

utkarshmani1997: We are returning the error to the callee, callee is printing the error : "operation not permitted". So printing the error here will pollute the logs unneccessary.

payes: Agreed

PR #108, opened 2018-08-29
 [TA2533] Remove peer details info

utkarshmani1997:
On branch US2682-quorum-fix
Changes to be committed:
	modified:   backend/file/file.go
	modified:   backend/remote/remote.go
	modified:   controller/client/controller_client.go
	modified:   controller/rebuild.go
	modified:   controller/replicator.go
	modified:   controller/rest/replica.go
	modified:   replica/peer_details.go
	modified:   replica/rest/model.go
	modified:   replica/rest/replica.go
	modified:   replica/rest/router.go
	modified:   replica/server.go
	modified:   sync/sync.go
	modified:   types/types.go

-  To remove the peer-details related code which was unused after the
   introduction of REPLICATION_FACTOR in the latest code.

-  Removed all the peer-details related code and their dependencies.

-  None

4. How to verify this change?
-  Run ci, it should work fine.

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -229,7 +229,7 @@ func (c *Controller) signalToAdd() {
 func (c *Controller) registerReplica(register types.RegReplica) error {
 	c.Lock()
 	defer c.Unlock()
-	logrus.Infof("Register Replica, Address: %v Uptime: %v State: %v Type: %v RevisionCount: %v PeerCount: %v",
+	logrus.Infof("Register Replica, Address: %v Uptime: %v State: %v Type: %v RevisionCount: %v",
 		register.Address, register.UpTime, register.RepState, register.RepType, register.RevCount)
 
 	_, ok := c.RegisteredReplicas[register.Address]


vishnuitta:
how is upgrade scenario taken care here? i.e., old replica with new controller or new controller with old replica.
@utkarshmani1997

utkarshmani1997:
@vishnuitta , one should ensure that there is no read , writes are happening. In upgrade scenarion either of the above case will work. for exp: have a look at this code given below.
```
package main

import (
	"encoding/json"
	"fmt"
	"net/http"
)

type User struct {
	Firstname string `json:"firstname"`
	Lastname  string `json:"lastname"`
}

func main() {
	http.HandleFunc("/decode", func(w http.ResponseWriter, r *http.Request) {
		var user User
		if err := json.NewDecoder(r.Body).Decode(&user); err != nil {
			fmt.Fprintf(w, "%v, %v", err, user)
		}

		fmt.Fprintf(w, "%s %s", user.Firstname, user.Lastname)
	})
	http.ListenAndServe(":8080", nil)
}
```
after running this program, run `curl -s -XPOST -d'{"firstname":"Utkarsh","lastname":"Tripathi"}' http://localhost:8080/decode` and you will get following output : 
`Utkarsh Tripathi`
try again with `curl -s -XPOST -d'{"firstname":"Utkarsh","lastname":"Tripathi", "location":"bangalore"}' http://localhost:8080/decode` and you will see the output is same, `Utkarsh Tripathi`

Same is true if reverse is happening. Conclusion is, it is purely based on json decoding which is done via json tags. So even if you send extra fields it will be ignored and if you send any of the missing fields default value of that field will be used. 

So either of the above case will work absolutely fine.

utkarshmani1997:
Closing this PR, Raised new PR #115

PR #110, opened 2018-08-31
fix(build): download qcow src from an active link

kmova:
jiva image is build by creating a build container that
downloads dependencies from external repositories. One
such dependency is on qcow tools project:
https://github.com/libyal/libqcow/

The older source tar has been removed from the above repo
which is causing jiva builds to fail. This PR updates
the build container to download from a recent active link.

Note: A more permanent solution is required to fix this flaky
builds due to breaking external dependencies. this will be taken 
up in the upcoming releases. 

Signed-off-by: kmova <kiran.mova@openebs.io>

### DIFF
--- a/Dockerfile.dapper
+++ b/Dockerfile.dapper
@@ -51,10 +51,10 @@ RUN cd /usr/src/tcmu-runner && \
     cp libtcmu_static.a /usr/local/lib/libtcmu.a
 
 # Install libqcow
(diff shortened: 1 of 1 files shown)


vishnuitta:
current code of jiva doesn't seem to use qcow backends. So, we should safe in removing this dependency and its files.
As mentioned in the description, this can be decided and implemented in upcoming release.

PR #111, opened 2018-09-03
[TA2643, TA2644] fix(snapshot): verify if snapshot already exist.

utkarshmani1997:
On branch US3005-verify-snapshot-name
Changes to be committed:

-  To fix the problem when snapshot with the given name already
   exist.

-  verify if snapshot already exist in the chain list and then
   proceed to create snapshot if snapshot with the given name
   does not exist.

-  None

4. How to verify this change?
-  Add ci for the same, it should run fine.

5. Any specific msg for the reviewer?
-  Refer to US3005
-  Snapshots are created in the linear order, need to optimize it in better way.


utkarshmani1997:
Closed, please refer to #564

PR #112, opened 2018-09-04
[TA2643, TA2644] fix(snapshot): verify if snapshot already exist

utkarshmani1997:
On branch US3005-verify-snapshot-name
Changes to be committed:

-  To fix the problem when snapshot with the given name already
   exist.

-  verify if snapshot already exist in the chain list and then
   proceed to create snapshot if snapshot with the given name
   does not exist.

-  None

4. How to verify this change?
-  Add ci for the same, it should run fine.

5. Any specific msg for the reviewer?
-  Refer to US3005
-  Snapshots are created in the linear order, need to optimize it in better way.

### DIFF
--- a/ci/start_init_test.sh
+++ b/ci/start_init_test.sh
@@ -837,10 +837,31 @@ run_data_integrity_test() {
 	#Cleanup is not being performed here because this data will be used to test clone feature in the next test
 }
 
-create_snapshot() {
-	echo "--------------create_snapshot-------------"
-	id=`curl http://$1:9501/v1/volumes | jq '.data[0].id' |  tr -d '"'`
-	curl -H "Content-Type: application/json" -X POST -d '{"name":"snap1"}' http://$CONTROLLER_IP:9501/v1/volumes/$id?action=snapshot
+# create_and_verify_snapshot first creates a snapshot and verifies if it
+# is of type snapshotOutput in case of failure it returns type:error which
+# is success case that the snapshot snap1 does not exists.
+# As a part of negative test we are trying to create the snapshot with the
+# same name which returns snapshot_exists.
+create_and_verify_snapshot() {
+	echo "--------------create_and_verify_snapshot-------------"
+       snapshot_exists="Snapshot: snap1 already exists"
+       id=`curl http://$CONTROLLER_IP:9501/v1/volumes | jq '.data[0].id' |  tr -d '"'`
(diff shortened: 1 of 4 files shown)


controller/control.go
```diff
@@ -334,6 +353,16 @@ func (c *Controller) Snapshot(name string) (string, error) {
 		return "", fmt.Errorf("Too many snapshots created")
 	}
 
+	for _, replica := range c.ListReplicas() {
```
payes: Why do we need to check at all the replicas.

ci/start_init_test.sh
```diff
@@ -837,10 +837,31 @@ run_data_integrity_test() {
...
+       fi;
+       message=`curl -H "Content-Type: application/json" -X POST -d '{"name":"snap1"}' http://$CONTROLLER_IP:9501/v1/volumes/$id?action=snapshot | jq '      .message' | tr -d '"'`
+       if [ "$message" == "$snapshot_exists" ];
+       then
+              echo "create and verify snapshot test2 passed"
+       else
+              echo "create and verify snapshot test2 failed"
+              collect_logs_and_exit
+       fi;
 }
```
payes: Let's create one more snapshot with a different name and check if something is breaking.
Lets also create, delete and recreate a snapshot with the same name and test.

utkarshmani1997: Thanks for the review. I have updated the PR with the above requested changes.

ci/start_init_test.sh
```diff
@@ -834,13 +831,49 @@ run_data_integrity_test() {
...
+# create_and_verify_snapshot creates a snapshot and verifies if it
+# has message given below (success case) or not (failure case).
+# As a part of negative test we are trying to create the snapshot with the
+# same name which returns snapshot_exists.
+create_and_verify_snapshot() {
+	snap='"snap'$2'"'
+	# remove trailing double quotes
+	snapname=$(echo $snap | tr -d '"')
+	message=`curl -H "Content-Type: application/json" -X POST -d '{"name":"snap'$2'"}' http://$CONTROLLER_IP:9501/v1/volumes/$1?action=snapshot | jq '.message' | tr -d '"'`
+	if [ "$message" == "Snapshot: $snapname created successfully" ] || [ "$message" == "Snapshot: $snapname already exists" ] ;
```
payes: In both successful and failure cases we are returning success.

utkarshmani1997: Thanks for pointing out this. Will update it.

PR #113, opened 2018-09-07
Remove old copies of data from auto-generated snapshots

payes:
[https://docs.google.com/document/d/1eBCIqq6dZ_8K8nHVkl82egPH3s7DIflv_FcPiATwehI/edit](url)


replica/backup.go
```diff
@@ -175,24 +180,92 @@ func (rb *Backup) findIndex(id string) int {
...
+
+var HoleCreatorChan = make(chan Hole, 1000000)
+
+//CreateHoles removes the offsets from corresponding sparse files
+func CreateHoles() error {
+	retryCount := 0
+	for {
+		hole := <-HoleCreatorChan
+	retry:
+		if err := syscall.Fallocate(int(hole.f.Fd()),
```
vishnuitta: When a diffDisk is used here, it should not get deleted from other ways like 'snap destroy' etc

replica/replica.go
```diff
@@ -255,6 +259,9 @@ func (r *Replica) insertBackingFile() {
 	d := disk{Name: r.info.BackingFile.Name}
 	r.activeDiskData = append([]*disk{{}, &d}, r.activeDiskData[1:]...)
 	r.volume.files = append([]types.DiffDisk{nil, r.info.BackingFile.Disk}, r.volume.files[1:]...)
+	r.volume.ReadOnlyIndx = append(r.volume.ReadOnlyIndx, false)
+	r.diskList = append(r.diskList, r.info.BackingFile.Name)
+	r.volume.UserCreatedSnap = append([]bool{false, false}, r.volume.UserCreatedSnap[1:]...)
```
vishnuitta: here, the append should be reverse

replica/backup.go
```diff
@@ -175,24 +180,92 @@ func (rb *Backup) findIndex(id string) int {
...
 			}
 		}
+		if d.UserCreatedSnap[i] {
+			userCreatedSnapIndx = byte(i)
+		}
 		generator := newGenerator(d, f)
 		for offset := range generator.Generate() {
+			if d.location[offset] != 0 {
+				if d.files[d.location[offset]] != file {
+					if file != nil && fileIndx > userCreatedSnapIndx && !d.ReadOnlyIndx[fileIndx] {
```
vishnuitta: file and fileIndx are not initialized, and in the loop, it will take previous loop contents.
So, after punching holes to 'file1', and, if any number of initial extents in next file i.e., 'file2' matches with 'file1', they won't get unmapped.
And, a comment would be helpful to understand code

replica/backup.go
```diff
@@ -175,24 +180,92 @@ func (rb *Backup) findIndex(id string) int {
...
+
+//CreateHoles removes the offsets from corresponding sparse files
+func CreateHoles() error {
+	retryCount := 0
+	for {
+		hole := <-HoleCreatorChan
+	retry:
+		if err := syscall.Fallocate(int(hole.f.Fd()),
+			sparse.FALLOC_FL_KEEP_SIZE|sparse.FALLOC_FL_PUNCH_HOLE,
+			hole.offset*defaultSectorSize, hole.len*4096); err != nil {
```
vishnuitta: here, offset and len should be multiplied by d.sectorsize.. let us pass sectorsize also in the Hole data structure or multiply it and add it to the channel

replica/backup.go
vishnuitta: got to understand that we are trying to find continuous blocks

vishnuitta: but, still, being file not initialized or initialized to NULL by default, it gets into 'else' case where 'length' is just increased.
Seems to be something wrong here.

vishnuitta: same file has been verified.. but, contiguity of blocks also need to be verified

vishnuitta: is it better to check of ReadOnlyIndx while doing the syscall for unmap?

replica/backup.go
```diff
@@ -186,13 +237,43 @@ func preload(d *diffDisk) error {
 				d.location[j] = 0
 			}
 		}
+		if d.UserCreatedSnap[i] {
+			userCreatedSnapIndx = byte(i)
+		}
 		generator := newGenerator(d, f)
 		for offset := range generator.Generate() {
+			if d.location[offset] != 0 {
+				//We are looking for continuous blocks over here.
```
vishnuitta: Add comment regarding the need for `userCreatedSnapIndx`

replica/replica.go
```diff
@@ -825,11 +851,20 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
...
+		r.diskList = append(r.diskList, newSnapName)
 	}
 	delete(r.diskData, oldHead)
+	//delete(r.diskList, oldHead)
 
 	r.info = info
 	r.volume.files = append(r.volume.files, f)
+	r.volume.ReadOnlyIndx = append(r.volume.ReadOnlyIndx, false)
+	r.diskList = append(r.diskList, newHeadDisk.Name)
+	if userCreated {
```
vishnuitta: why `- 2`? userCreated snapshot will be just next to the head right? so, should be `- 1`?

vishnuitta: do we need to lock protect this?

replica/replica.go
payes: It is sync with r.volume.files. Although this feature is not being used currently.

replica/diff_disk.go
```diff
@@ -182,8 +227,12 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 		if newTarget == target {
 			readSectors++
 		} else {
-			c, err := d.read(target, buf, offset, i-readSectors, readSectors)
-			count += c
+			if target == 0 {
```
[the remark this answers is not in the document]
payes: Target gives the index of the file in which that offset is latest. If target=0 implies the data has never been written on that offset, we can simply return zeroes. File indexes start from 1.

replica/replica.go
payes: Lock protection is not required as all the functions which are using it are lock protected.

replica/backup.go
payes: This will be covered in snap destroy PR.

vishnuitta: as we don't delete internal snapshots, it should be fine.

replica/backup.go
payes: This has been replaced with already present snapIndx.

PR #114, opened 2018-09-12
[ON-HOLD] feat(jiva): implement delete command

utkarshmani1997:
Implement delete command in jiva to purge replica files.
exp: `longhorn del`
Refactor the delete api.
Introduce various error codes for the errors.

On branch US3115-refactor-delete-api
Changes to be committed:
	new file:   app/del_replica.go
	modified:   controller/client/controller_client.go
	modified:   controller/rest/delete.go
	modified:   controller/rest/router.go
	modified:   controller/rest/volume.go
	modified:   main.go

Output:
```
failure case: status code: 404
curl -X "DELETE" 172.18.0.2:9501/v1/volumes/c3RvcmUx?action=deletevolume

{
  "actions":{},
  "links": {
    "self":"http://172.18.0.2:9501/v1/volumes/c3RvcmUx"
  },
  "replicas": [
    {
      "error": "Volume not found",
      "msg": "No replicas registered with this controller instance",
      "replica":""
    }
  ],
  "type": "deleteVolumeOutput"
}

success case: status code: 200

{
  "actions": {},
  "links": {
    "self": "http://172.18.0.2:9501/v1/volumes/c3RvcmUx"
  },
  "replicas": [
    {
      "msg": "In Progress",
      "replica": "tcp://172.18.0.4:9502"
    },
    {
      "msg": "In Progress",
      "replica": "tcp://172.18.0.3:9502"
    }
  ],
  "type": "deleteVolumeOutput"
}

failure case: status code: 409

{
  "actions":{},
  "links": {
    "self":"http://172.18.0.2:9501/v1/volumes/c3RvcmUx"
  },
  "replicas": [
    {
      "error": "Replication factor: 2 is not equal to replica count: 1",
      "msg": "Error deleting replica",
      "replica":""
    }
  ],
  "type": "deleteVolumeOutput"
}

and status code : 500 for other errors
```

### DIFF
--- a/controller/rest/delete.go
+++ b/controller/rest/delete.go
@@ -14,11 +14,10 @@ import (
 )
 
 const (
-	volumeNotFound  = "volume not found"
-	zeroReplica     = "No replicas registered with this controller instance"
-	repClientErr    = "error in creating replica client"
-	deletionSuccess = "Replica deleted successfully"
-	deletionErr     = "Error deleting replica"
+	volumeNotFound = "volume not found"
+	zeroReplica    = "No replicas registered with this controller instance"
+	repClientErr   = "error in creating replica client"
+	deletionErr    = "Error deleting replica"
 )
 
 type DeletedReplica struct {
@@ -65,12 +64,13 @@ func (s *Server) delete(replicas *DeletedReplicas, wg *sync.WaitGroup) {
 				return
 			}
 			logrus.Info("Sending delete request to replica : ", addr)
-			if err := repClient.Delete("/delete"); err != nil {
+			status, err := repClient.Delete("/delete")
+			if err != nil {
 				logrus.Infof("Error in delete operation of replica %v , error %v", addr, err)
 				replicas.appendDeletedReplicas(err.Error(), addr, deletionErr)
 				return
 			}
-			replicas.appendDeletedReplicas("", addr, deletionSuccess)
+			replicas.appendDeletedReplicas("", addr, status)
 		}(addr)
 	}
 }
--- a/replica/client/client.go
+++ b/replica/client/client.go
@@ -77,21 +77,29 @@ func (c *ReplicaClient) Create(size string) error {
 	}, nil)
 }
 
-func (c *ReplicaClient) Delete(path string) error {
+func (c *ReplicaClient) Delete(path string) (string, error) {
+	var resp rest.DeleteReplicaOutput
 	_, err := c.GetReplica()
 	if err != nil {
-		return err
+		return "", err
 	}
 	deleteAPI := c.address + path
 	req, err := http.NewRequest("DELETE", deleteAPI, nil)
 	if err != nil {
-		return err
+		return "", err
 	}
-	_, err = c.httpClient.Do(req)
+	httpResp, err := c.httpClient.Do(req)
 	if err != nil {
-		return err
+		return "", err
 	}
-	return nil
+	if httpResp.StatusCode >= 300 {
+		content, _ := ioutil.ReadAll(httpResp.Body)
+		return "", fmt.Errorf("Failed to delete replica contents: %d %s: %s", httpResp.StatusCode, httpResp.Status, content)
+	}
+	if err = json.NewDecoder(httpResp.Body).Decode(&resp); err != nil {
+		return "", err
+	}
+	return resp.Status, err
 }
 
 func (c *ReplicaClient) Revert(name, created string) error {
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -1010,15 +1010,16 @@ func (r *Replica) Delete() error {
 	return nil
 }
 
-func (r *Replica) DeleteAll() error {
+func (r *Replica) DeleteAll() {
 	r.Lock()
 	defer r.Unlock()
 
-	if err := os.RemoveAll(r.dir); err != nil {
-		logrus.Error("Error in deleting the directory contents, error : ", err.Error())
-		return err
-	}
-	return nil
+	go func() {
+		if err := os.RemoveAll(r.dir); err != nil {
+			logrus.Error("Error in deleting the directory contents, error : ", err.Error())
+		}
+	}()
+	return
 }
 
 func (r *Replica) Snapshot(name string, userCreated bool, created string) error {
--- a/replica/rest/delete.go
+++ b/replica/rest/delete.go
@@ -22,6 +22,7 @@ func (s *Server) DeleteVolume(rw http.ResponseWriter, req *http.Request) error {
 		client.Resource{
 			Type: "delete",
 		},
+		"In Progress",
 	})
 	return nil
 }
--- a/replica/rest/model.go
+++ b/replica/rest/model.go
@@ -29,6 +29,7 @@ type Replica struct {
 
 type DeleteReplicaOutput struct {
 	client.Resource
+	Status string `json:"status"`
 }
 
 type Stats struct {
--- a/replica/server.go
+++ b/replica/server.go
@@ -360,9 +360,9 @@ func (s *Server) DeleteAll() error {
 	}
 
 	logrus.Infof("Deleting all the contents of the volume")
-	err = s.r.DeleteAll()
+	s.r.DeleteAll()
 	s.r = nil
-	return err
+	return nil
 }
 
 func (s *Server) Close(signalMonitor bool) error {


app/del_replica.go
```diff
@@ -0,0 +1,38 @@
...
+	"text/tabwriter"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+)
+
+func DelReplicaCmd() cli.Command {
+	return cli.Command{
+		Name:      "del-replica",
+		ShortName: "del",
```
gila: Does this mean that I type mayactl del-replica or mayactl del ? 

Why not mayactl delete <replica | volume> < replicaname | volumename> ?

app/del_replica.go
```diff
@@ -0,0 +1,38 @@
...
+			if err := deleteReplica(c); err != nil {
+				logrus.Fatalf("Error running del replica command: %v", err)
+			}
+		},
+	}
+}
+
+func deleteReplica(c *cli.Context) error {
+	controllerClient := getCli(c)
+	out, err := controllerClient.DeleteVolume()
```
gila: Delete replica vs. delete volume? Which is it?

controller/client/controller_client.go
```diff
@@ -116,6 +116,33 @@ func (c *ControllerClient) CreateQuorumReplica(address string) (*rest.Replica, e
...
+	}
+	httpReq, err := http.NewRequest("DELETE", vol.Actions["deletevolume"], nil)
+	if err != nil {
+		return nil, err
+	}
+	httpResp, err := http.DefaultClient.Do(httpReq)
+	if err != nil {
+		return nil, err
+	}
+	if httpResp.StatusCode >= 300 {
```
gila: Wait... >= 300 is a bad response? How so? What errors can we get back, to begin with? Also is this method synchronous? What if the volume is large? Would it not time out? Should we not implement this?

controller/rest/delete.go
```diff
@@ -84,18 +85,29 @@ func (s *Server) DeleteVolume(rw http.ResponseWriter, req *http.Request) error {
 		replicas DeletedReplicas
 		wg       sync.WaitGroup
 	)
+	logrus.Info("Delete volume")
 	apiContext := api.GetApiContext(req)
+	id := mux.Vars(req)["id"]
+	v := s.getVolume(apiContext, id)
+	if v == nil {
+		rw.WriteHeader(http.StatusNotFound)
```
gila: So here we set a 404 (not found) but in the above code we will say its a bad request. Its *not* a bad request -- the volume was simply not found

controller/rest/delete.go
```diff
@@ -84,18 +85,29 @@ func (s *Server) DeleteVolume(rw http.ResponseWriter, req *http.Request) error {
...
+		return fmt.Errorf("%v", volumeNotFound)
+	}
 	replicaCount := len(s.c.ListReplicas())
 	if replicaCount == 0 {
```
gila: What does it mean to have a volume with zero replica's?

controller/rest/model.go
```diff
@@ -138,6 +138,7 @@ func NewVolume(context *api.ApiContext, name string, readOnly bool, replicas int
 		v.Actions["shutdown"] = context.UrlBuilder.ActionLink(v.Resource, "shutdown")
 		v.Actions["snapshot"] = context.UrlBuilder.ActionLink(v.Resource, "snapshot")
 		v.Actions["revert"] = context.UrlBuilder.ActionLink(v.Resource, "revert")
+		v.Actions["deletevolume"] = context.UrlBuilder.ActionLink(v.Resource, "deletevolume")
```
gila: As its possible to have a volume with 0 replica's this should be delete replica's?

app/del_replica.go
utkarshmani1997: no this is `longhorn del` and it  is implemented based on their existing implementation of the other commands.

app/del_replica.go
utkarshmani1997: it's delete volume. Will change that.

controller/client/controller_client.go
utkarshmani1997: As per the handling we will only get, 200, 404(volume not found), 409(Replication factor conflict) and 500(Bad response from replica), So we should handle this case to reply the client about the cause of error. 
Regarding the timeout I haven't tested it on the larger volumes. It uses SYS_UNLINKAT internally, i guess it mark the file as delete and comes out and deletion of the contents happens parallelly across replicas.

controller/rest/delete.go
utkarshmani1997: Correct, will fix the log.

controller/rest/delete.go
utkarshmani1997: It's possible that volume structure is not updated which contains the various links related to the operations such as shutdown, start and rebuild. So to reverify the events checking the replica count. Should not we check replica count ?

controller/rest/model.go
utkarshmani1997: We are deleting the whole volume and there is already deletereplica implemented which only remove the metadata of the replica not the contents in that directory. So i think deletevolume sounds good. What do you say ?

ci/start_init_test.sh
```diff
@@ -395,7 +395,8 @@ verify_delete_replica_unsuccess() {
 verify_delete_replica() {
     old_replica_count=$(get_replica_count $CONTROLLER_IP)
     echo "$old_replica_count"
-    curl -X "POST" http://$CONTROLLER_IP:9501/v1/delete | jq
+    id=`curl http://$CONTROLLER_IP:9501/v1/volumes | jq '.data[0].id' |  tr -d '"'`
+    curl -X "DELETE" http://$CONTROLLER_IP:9501/v1/volumes/$id?action=deletevolume | jq
```
princerachit: Would it make sense to update the appropriate method in controller wo that we don't need to pass `action=deletevolume`? The Header `DELETE` becomes redundant.

utkarshmani1997: @princerachit i'm not getting you.

app/del_replica.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/codegangsta/cli"
+)
+
+func DelReplicaCmd() cli.Command {
+	return cli.Command{
+		Name:      "del-replica",
+		ShortName: "del",
+		Action: func(c *cli.Context) {
+			if err := deleteReplica(c); err != nil {
+				logrus.Fatalf("Error running del replica command: %v", err)
```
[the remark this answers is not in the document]
princerachit: Yes that was my question. Why do we need to crash?

utkarshmani1997:
Thanks @gila , @princerachit for the review. Will update the PR soon

ci/start_init_test.sh
princerachit: Why do we need action delete when we are requesting with header DELETE?

app/del_replica.go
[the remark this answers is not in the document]
utkarshmani1997: because we don't need to let it continue.

ci/start_init_test.sh
utkarshmani1997: We generate the link for the set of actions such as revert, shutdown, start etc. So we get a link for delete action. I believe having an link for the delete action ensures we are performing right operation. Now question is what if we don't make use of action at all.

It's possible to do so, but this allows us to perform delete only with one call but above approach helps us to validate the delete url first (by generating a valid url which has been done earlier) and then perform delete.

controller/client/controller_client.go
gila: So how is the user to understand, the differences between 404 and lets sat 409? 

As for deleting "large" files -- we would need to implement this async. So we post the request to delete it, get back a 202 (from memory, we should double check that) and use the resulting URL, to poll the status of the request.

utkarshmani1997: One of the main challenge is, if we start deleting the replica all the fields are set to nil, So replica IP will be removed from the controller. So if someone wants to see the status of the volume they can get at `REPLICA_IP:9502/v1/delete`
Following are the status:
"In Progress"
"Completed"
"(Error msg in case of any error)"

So it will reply the above msg's for the delete call for the first request based on the status of replica at that point of time.

gila: I'm not sure I understand what you are trying to say, sorry my bad. Not sure if you understand my comment either, but to hopefully make less unclear. 

If we are accepting a delete request, which we process in the background, the server -- should have a queue of work in progress that the client can poll. This is a pretty common approach in rest land. The response object of the request contains an URL, which the callee can poll until it completes (or fails).

I hope this makes sense to you

utkarshmani1997: @gila i going with async approach where it will start deleting the contents of the volume in the background and it will return "InProgress" immediately. So even if someone makes a delete call again, it will show "resourse is busy" (if deletion is still happening in the background) else it will return error saying "s.r = nil".
We will implement at the delete operation via polling later in some other PR.

gila: That's sounds OK 👍

kmova:
This has been handled from maya by launching a k8s job to clean up the replica data directory. https://github.com/openebs/maya/pull/786

PR #116, opened 2018-09-15
[TA2880, TA2881] fix(jiva): extents mapping bug

utkarshmani1997:
On branch US3116-extent-mapping-bug
Changes to be committed:
	modified:   replica/server.go

This commit fixes the bug which was not fataling the replicas
if they are restarted due to unsupported underlying file system
where replicas store the data.

Add ci to verify the fix.

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -163,6 +163,9 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		logrus.Errorf("failed to create directory: %s", dir)
 		return nil, err
 	}
+	if err := findExtents(dir); err != nil {
+		return nil, err
+	}
 
 	r := &Replica{
 		dir:             dir,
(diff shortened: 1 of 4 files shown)


replica/replica.go
```diff
@@ -225,8 +228,8 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.ReplicaType = replicaType
 
 	if err := PreloadLunMap(&r.volume); err != nil {
-		logrus.Error("underlying file system does not support extent mapping")
-		return r, err
+		r.Delete()
```
payes: This call is not required as PreloadLunMap can fail due to multiple issues.

replica/replica.go
```diff
@@ -163,6 +163,9 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		logrus.Errorf("failed to create directory: %s", dir)
 		return nil, err
 	}
+	if err := findExtents(dir); err != nil {
```
payes: This should be checked just once while starting replica but Construct function is called multiple times.
Lets also give a better name for this function, like isExtentsSupported, or something else.

PR #117, opened 2018-09-19
[TA3086] fix(replica): don't add replica if RF is met

utkarshmani1997:
On branch 3006-fix-add-replica
Changes to be committed:
	modified:   controller/rest/replica.go

This commit fix the issue where the new replicas are getting added
even if the RF condition is met. i.e, (Replication Factor = Replica
count)


utkarshmani1997:
Thanks @payes for the review. Added the ci for the same.

PR #118, opened 2018-09-21
[TA2752, TA2754] feat(jiva): delete snapshot api

utkarshmani1997:
Changes to be committed:
	modified:   controller/rest/delete.go
	modified:   controller/rest/router.go
	modified:   controller/rest/volume.go

Add api to delete the user created snapshots.

Deletion doesn't work if there exist only one snapshot.
Figuring out the way to delete the snapshot. Please
donot merge, work in progress.


utkarshmani1997:
@payes in golang they don't allow cyclic imports, so here three packages call each other in cyclic way
This is due to the rm snapshot command. So it is very difficult to decouple them. The scenario is given below. So it will require a lot of work. If we are good with it then put this PR on hold and i will rework on it.
                        app/snapshot.go 
                       ^                          \
                     /                              \
                   /                                 v
controller/rest/delete.go < - - - - -  sync/sync.go

utkarshmani1997:
@payes  i have raised a new fresh PR #153 , so closing this.

PR #121, opened 2018-10-01
[US3272] fix(resize): Fix replica crash due to resize

payes:
Location table contains a mapping of the block number of the volume to the snapshot file containing its latest copy.
After resizing the volume at replica, we were not increasing the size of the location table.
When a read/write was being received for that block, we were trying to access the same location in the table which was causing the crash.
  
### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -296,6 +296,8 @@ func (r *Replica) Resize(obj interface{}) error {
 			return nil
 		}
 	}
+	r.Lock()
+	defer r.Unlock()
 	if r.info.Size > sizeInBytes {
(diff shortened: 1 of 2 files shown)


replica/replica.go
```diff
@@ -304,6 +306,8 @@ func (r *Replica) Resize(obj interface{}) error {
 			return err
 		}
 	}
+	byteArray := make([]byte, (sizeInBytes-r.info.Size)/4096)
```
vishnuitta: what if r.info.Size is > sizeInBytes?

payes: This is already handled at the controller from where we get this rest call. But its good to add a check here also. I will add it. Thanks @vishnuitta .

PR #24, opened 2017-10-07
make docker image smaller

circa10a:
optimize docker image, reference openebs/openebs#553

PR #128, opened 2018-10-09
(issue-#24) enhancing performace avoiding defer unlocks

khushbuparakh:
Signed-off-by: khushbuparakh <khushbuparakh@hotmail.com>
I started from ReadAt, WriteAt functions.
So func (s *Server) had a defer func.
I avoided using that defer function
Then used at the very low level I/O code and
follow the path to know IO path in controller side, and,
similarly Handle() in rpc/server.go for replica side IO handling.

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -100,12 +100,12 @@ func (r *Remote) doAction(action string, obj interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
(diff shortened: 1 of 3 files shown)


khushbuparakh:
yeah! @payes fixing it

utkarshmani1997:
LGTM

utkarshmani1997:
Thanks @khushbuparakh for the PR. With the latest version of go there is significant improvement in performance with the use of defer as mentioned in this [issue](https://github.com/golang/go/issues/14939). So closing this PR

PR #142, opened 2018-10-16
updated gitlab-ci yml and push script to build and push commit based

atulabhi:
Commit based images are required to run OpenEBS-E2E . These images will be push by gitlab-runner based on the commit from the github. The commit is fetch from github and update as an environment variable (COMMIT ) in gitlab-runner which is being utilised in the push script.

### DIFF
--- a/scripts/push
+++ b/scripts/push
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -ex
+set -e
 
 if [ -z ${IMAGE_REPO} ];
 then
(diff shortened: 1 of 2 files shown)


scripts/push
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -e
+set -ex
```
utkarshmani1997: remove `x` , this will print the value of DNAME and DPASS

PR #153, opened 2018-10-26
[TA2752, TA2754] feat(jiva): delete snapshot api

utkarshmani1997:
Add api to delete the user created snapshots.

Deletion doesn't work if there exist latest snapshot because its
contains the latest data and it will be marked as removed.


utkarshmani1997:
closing this via #232

PR #159, opened 2018-11-05
Cherry-pick commits from master

payes:


app/replica.go
```diff
@@ -187,8 +187,8 @@ func startReplica(c *cli.Context) error {
 		server := rest.NewServer(s)
 		router := http.Handler(rest.NewRouter(server))
 		router = util.FilteredLoggingHandler(map[string]struct{}{
-			"/ping":          struct{}{},
```
vishnuitta: any upgrade scenarios we need to worry about? like some modules are in 0.6 or 0.7, and, replica is in 0.7.2 or other way round

PR #161, opened 2018-11-12
[TA3939, TA3971] feat(stats): Add rep/cont status field

utkarshmani1997:
This commit adds the replica and controller status field in the stats
api
fix the issue in revision counter, where it was giving zero value
Add json tags into Replica struct

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -176,6 +176,7 @@ func (r *Remote) GetVolUsage() (types.VolUsage, error) {
 	}
 
 	err = json.NewDecoder(resp.Body).Decode(&Details)
+	volUsage.RevisionCounter, _ = strconv.ParseInt(Details.RevisionCounter, 10, 64)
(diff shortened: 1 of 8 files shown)


backend/remote/remote.go
```diff
@@ -176,6 +176,7 @@ func (r *Remote) GetVolUsage() (types.VolUsage, error) {
 	}
 
 	err = json.NewDecoder(resp.Body).Decode(&Details)
+	volUsage.RevisionCounter, _ = strconv.ParseInt(Details.RevisionCounter, 10, 64)
```
ashishranjan738: shouldn't we handle the error from strconv ?

utkarshmani1997: @ashishranjan738 there are only two cases when it fails
1. If the length is > len(uint64/int64) or if the length is zero
for both the case we want the output to be zero, that's why not caching the error.

PR #163, opened 2018-11-19
Add support for Sync and Unmap

payes:

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -831,6 +831,54 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
+func (c *Controller) Sync() (err error) {
+	c.Lock()
+	if c.ReadOnly == true {
+		err := fmt.Errorf("Mode: ReadOnly")
+		c.Unlock()
+		time.Sleep(1 * time.Second)
+		return err
+	}
+	defer c.Unlock()
+	err = c.backend.Sync()
+	if err != nil {
+		errh := c.handleErrorNoLock(err)
+		if bErr, ok := err.(*BackendError); ok {
+			if len(bErr.Errors) > 0 {
+				for address := range bErr.Errors {
+					c.RemoveReplicaNoLock(address)
+				}
+			}
+		}
+		return errh
+	}
+	return err
+}
+
+func (c *Controller) Unmap(offset int64, length int) (err error) {
+	c.Lock()
+	if c.ReadOnly == true {
+		err := fmt.Errorf("Mode: ReadOnly")
+		c.Unlock()
+		time.Sleep(1 * time.Second)
+		return err
+	}
+	defer c.Unlock()
+	err = c.backend.Unmap(offset, length)
+	if err != nil {
+		errh := c.handleErrorNoLock(err)
+		if bErr, ok := err.(*BackendError); ok {
+			if len(bErr.Errors) > 0 {
+				for address := range bErr.Errors {
+					c.RemoveReplicaNoLock(address)
+				}
+			}
+		}
+		return errh
+	}
+	return err
+}
+
 func (c *Controller) ReadAt(b []byte, off int64) (int, error) {
 	c.Lock()
 	defer c.Unlock()
--- a/controller/control.go
+++ b/controller/control.go
@@ -831,16 +831,16 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	return n, err
 }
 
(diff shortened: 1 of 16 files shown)


rpc/client.go
```diff
@@ -18,6 +18,8 @@ var (
 	opRetries       = 0
 	opReadTimeout   = 15 * time.Second // client read
 	opWriteTimeout  = 15 * time.Second // client write
+	opSyncTimeout   = 30 * time.Second // client sync
+	opUnmapTimeout  = 30 * time.Second // client unmap
```
utkarshmani1997: What will happen if the io's are larger in size ? Sync and unmap may fail intermittently due to timeouts.

types/types.go
```diff
@@ -130,5 +136,7 @@ type Frontend interface {
 type DataProcessor interface {
 	ReaderWriterAt
 	PingResponse() error
+	Sync() error
+	Unmap(int64, uint32) error
 	//Update() error
```
utkarshmani1997: Since sync and unmap is used in many packages like types , replica , frontend etc. Can we write reason what is the need of this in every package, since it will be very difficult for the new comer to understand why the same function is being used everywhere ?

replica/diff_disk.go
```diff
@@ -108,6 +110,22 @@ func (d *diffDisk) readModifyWrite(buf []byte, offset int64) (int, error) {
...
 }
 
+func (d *diffDisk) Sync() (err error) {
+	target := byte(len(d.files) - 1)
+	fd := d.files[target].Fd()
+	err = syscall.Fsync(int(fd))
+	return
+}
+
+func (d *diffDisk) Unmap(offset int64, length uint32) error {
```
utkarshmani1997: can we put some comments about what unmap is doing ?

controller/control.go
```diff
@@ -831,6 +831,54 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
...
+	if err != nil {
+		errh := c.handleErrorNoLock(err)
+		if bErr, ok := err.(*BackendError); ok {
+			if len(bErr.Errors) > 0 {
+				for address := range bErr.Errors {
+					c.RemoveReplicaNoLock(address)
+				}
+			}
+		}
+		return errh
```
vishnuitta: here, this may return nil when c.backend.Sync returned errors with more than (m.writers/2), but, a replica being in rw, errh becomes nil. Need to handle this case.

controller/multi_writer_at.go
```diff
@@ -110,3 +110,91 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
...
+	//If any replica has errored, return with the length of data written and the
+	//erroed replica details so that it can be closed.
+	if replicaErrored {
+		for _, err1 := range replicaErrs {
+			if err1 != nil {
+				replicaErrCount++
+			}
+		}
+		if len(m.writers)-replicaErrCount > len(m.writers)/2 {
+			return &errors
```
vishnuitta: better to have different return values in case of partial success for 'having required success_cnt' and 'not having required success_cnt'

controller/control.go
vishnuitta: same can apply for unmap as well

replica/diff_disk.go
vishnuitta: unmap may need to be sent to all the diffDisks

controller/multi_writer_at.go
```diff
@@ -110,3 +110,91 @@ func (m *MultiWriterAt) WriteAt(p []byte, off int64) (n int, err error) {
...
+	wg := sync.WaitGroup{}
+	var errors MultiWriterError
+	var multiWriterMtx sync.Mutex
+
+	for i, w := range m.writers {
+		wg.Add(1)
+		go func(index int, w Writer) {
+			err := w.Sync()
+			if err != nil {
+				multiWriterMtx.Lock()
```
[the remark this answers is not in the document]
utkarshmani1997: Ok, since replicaErrored is being accessed concurrently, so this comment does not hold valid. Please ignore this

rpc/types.go
```diff
@@ -29,7 +31,7 @@ type Message struct {
 	Type         uint32
 	Offset       int64
 	Data         []byte
-	Size         uint32
+	Size         int64
```
vishnuitta: we need to increase MagicVersion?

vishnuitta: how about uint64?

PR #168, opened 2018-12-06
[TA4284] fix(rpc): close connection at the client side

utkarshmani1997:
Controller initiates a new connection upon replica registration,
this might create some problem due to concurrent goroutines, as some
of the goroutine may not have been exited. To ensure graceful close
of the connection, i have added CloseRead and CloseWrite in both 
replica and controller.

### DIFF
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -214,16 +214,35 @@ func (c *Client) operation(op uint32, buf []byte, offset int64, length int64) (i
 	}
 }
 
-/*
 //Close replica client
-func (c *Client) Close() {
-	c.wire.Close()
-	c.end <- struct{}{}
+func (c *Client) Close() error {
+	//	c.wire.Close()
+	//	c.end <- struct{}{}
+	if err := c.wire.CloseRead(); err != nil {
+		return err
+	}
+
+	if err := c.wire.CloseWrite(); err != nil {
+		return err
+	}
+	for {
+		if c.wire.readExit && c.wire.writeExit {
+			break
+		}
+		time.Sleep(2 * time.Second)
+	}
+	close(c.send)
+	return c.wire.Close()
 }
-*/
 
 func (c *Client) loop() {
-	defer close(c.send)
+	defer func() {
+	retry:
+		if err := c.Close(); err != nil {
+			logrus.Error("failed to close conn, err: ", err)
+			goto retry
+		}
+	}()
 
 	for {
 		select {
@@ -235,9 +254,6 @@ func (c *Client) loop() {
 			c.handleResponse(resp)
 			if c.err != nil {
 				logrus.Infof("Exiting rpc loop for %v with err %v", c.peerAddr, c.err)
-				if err := c.wire.Close(); err != nil {
-					logrus.Error("failed to close conn, err: ", err)
-				}
 				return
 			}
 		}
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -216,8 +216,6 @@ func (c *Client) operation(op uint32, buf []byte, offset int64, length int64) (i
 
 //Close replica client
 func (c *Client) Close() error {
-	//	c.wire.Close()
-	//	c.end <- struct{}{}
 	if err := c.wire.CloseRead(); err != nil {
 		return err
 	}
@@ -244,8 +242,6 @@ func (c *Client) loop() {
 
 	for {
 		select {
-		//		case <-c.end:
-		//			return
 		case req := <-c.requests:
 			c.handleRequest(req)
 		case resp := <-c.responses:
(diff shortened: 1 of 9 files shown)


rpc/client.go
```diff
@@ -232,6 +232,15 @@ func (c *Client) loop() {
 			c.handleResponse(resp)
 			if c.err != nil {
 				logrus.Infof("Exiting rpc loop for %v with err %v", c.peerAddr, c.err)
+				if err := c.wire.CloseRead(); err != nil {
+					logrus.Error("failed to close read, error: ", err)
+				}
+				if err := c.wire.CloseWrite(); err != nil {
+					logrus.Error("failed to close write, error: ", err)
+				}
+				if err := c.wire.Close(); err != nil {
```
vishnuitta: We can't be sure that other GO routines are exited. So, doing Close here can be risky as the same fd can be re-assigned when other GO routines are still using it.

vishnuitta:
Please try code as below to make sure GO routines are exited:
```
diff --git a/rpc/client.go b/rpc/client.go
index 6fc1a75..6fb8ce8 100644
--- a/rpc/client.go
+++ b/rpc/client.go
@@ -70,6 +70,8 @@ func NewClient(conn net.Conn, closeChan chan struct{}) *Client {
                messages:  map[uint32]*Message{},
                closeChan: closeChan,
        }
+       c.write_exited = false
+       c.read_exited = false
        go c.loop()
        go c.write()
        go c.read()
@@ -221,6 +223,17 @@ func (c *Client) Close() {
 
 func (c *Client) loop() {
        defer close(c.send)
+       defer func() {
+               c.wire.CloseRead()
+               c.wire.CloseWrite()
+               for {
+                       if (c.read_exited == true) && (c.write_exited == true) {
+                               break
+                       }
+                       time.Sleep()
+               }
+               c.wire.Close()
+       }
 
        for {
                select {
@@ -317,6 +330,7 @@ func (c *Client) write() {
                }
        }
        logrus.Infof("Exiting rpc writer, RemoteAddr:%v", c.peerAddr)
+       c.write_exited = true
 }
 
 func (c *Client) read() {
@@ -330,4 +344,5 @@ func (c *Client) read() {
                c.responses <- msg
        }
        logrus.Infof("Exiting rpc reader, RemoteAddr:%v", c.peerAddr)
+       c.read_exited = true
 }
```

rpc/server.go
```diff
@@ -42,20 +45,33 @@ func (s *Server) Handle() error {
 	}()
 	ret := make(chan error)
 	go s.readWrite(ret)
-
+	s.pingStart = time.Now()
```
vishnuitta: lets move this to above s.readWrite

rpc/server.go
```diff
@@ -42,26 +46,32 @@ func (s *Server) Handle() error {
...
+				if err := s.Stop(); err != nil {
+					return err
+				}
+			}
 		}
 	}
 }
 
 func (s *Server) readWrite(ret chan<- error) {
+	s.pingStart = time.Now()
```
vishnuitta: move this to `Handle` function

rpc/server.go
```diff
@@ -3,17 +3,21 @@ package rpc
...
-	responses   chan *Message
-	done        chan struct{}
-	data        types.DataProcessor
-	monitorChan chan struct{}
+	wire                *Wire
+	responses           chan *Message
+	done                chan struct{}
+	data                types.DataProcessor
+	monitorChan         chan struct{}
+	pingStart           time.Time
```
vishnuitta: change this to `pingRecvd` and add comments for the same

rpc/server.go
```diff
@@ -88,15 +98,35 @@ func (s *Server) readWrite(ret chan<- error) {
...
+	s.writeExit = true
 }
 
-func (s *Server) Stop() {
-	s.done <- struct{}{}
+func (s *Server) Stop() error {
+	if err := s.wire.CloseRead(); err != nil {
+		return err
+	}
+	s.readExit = true
```
vishnuitta: readExit is not required. And, change, writeExit to readWriteExit

rpc/wire.go
```diff
@@ -54,6 +55,9 @@ func (w *Wire) Write(msg *Message) error {
 		logrus.Errorf("Write len(msg.Data) failed, Error: %v", err)
 		return err
 	}
+	if w.readExit {
```
vishnuitta: remove this check. Not required.

rpc/wire.go
```diff
@@ -101,7 +105,9 @@ func (w *Wire) Read() (*Message, error) {
 		logrus.Errorf("Read length failed, Error: %v", err)
 		return nil, err
 	}
-
+	if w.writeExit {
```
vishnuitta: remove this check. Not required.

PR #170, opened 2018-12-18
[TA4197] fix(replica): preload issue due to large no of extents

utkarshmani1997:
Recently we have seen that if volume size is large then preload
takes time and hence openReplica fails due to timeouts and hence
neither of them succedes.

There are five calls where Preload is required
- When replica is in Initial state
- Replica received `open` signal from controller
- Replica is closed
- Replica is reloading
- Revert to snapshot

To solve the issue, we have removed it from the construct function
rather we are calling it after the sync is done.

Feature : Added `PreloadStatus` field in Replica structure, so status can be queried via REST api
There are three status:
```
Error    : In case of replica disconnection or syscall failure
Started  : If Preload is happening in the background
Done     : If Preload is completed. 
```
NOTE:
- Since we are running IO's and preload parallely, we may miss some block counts for iostats, this can be taken as improvement for the next iteration

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -86,7 +86,7 @@ func (c *Controller) UpdateVolStatus() {
 	} else {
 		c.ReadOnly = true
 	}
-	logrus.Infof("controller readonly p:%v c:%v rcount:%v rw_count:%v",
+	logrus.Infof("Controller's previous RO status: %v, current: %v replica count:%v,  RW replica count:%v",
 		prev, c.ReadOnly, len(c.replicas), rwReplicaCount)
 	logrus.Infof("backends len: %d", len(c.backend.backends))
 }
@@ -706,7 +706,8 @@ getCloneStatus:
 	} else if status == "error" {
 		return fmt.Errorf("Replica clone status returned error %s", address)
 	}
-	c.setReplicaModeNoLock(address, types.RW)
+	c.setReplicaModeNoLock(address, types.WO)
+	go c.monitorPreload(address)
 	return nil
 }
 
@@ -730,8 +731,6 @@ func (c *Controller) Start(addresses ...string) error {
 	}
 
 	c.reset()
-
-	defer c.startFrontend()
 
 	c.size = math.MaxInt64
 	for _, address := range addresses {
@@ -790,7 +789,6 @@ func (c *Controller) Start(addresses ...string) error {
 	}
 	logrus.Info("Update volume status")
 	c.UpdateVolStatus()
-
 	return nil
 }
 
@@ -962,9 +960,11 @@ func (c *Controller) shutdownFrontend() error {
 	c.RLock()
 	defer c.RUnlock()
 
+	logrus.Info("Stopping controller")
 	if c.frontend != nil {
 		return c.frontend.Shutdown()
 	}
+
 	return nil
 }
 
@@ -1034,4 +1034,15 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
 	}
 	logrus.Infof("Monitoring stopped %v", address)
 	c.RemoveReplicaNoLock(address)
+}
+
+func (c *Controller) monitorPreload(address string) {
+	logrus.Infof("Get preload status of backend %s", address)
+	if err := c.backend.GetPreloadStatus(address); err != nil {
+		logrus.Errorf("Failed to monitor preload, error: %v", err)
+		return
+	}
+	c.startFrontend()
+	c.setReplicaModeNoLock(address, types.RW)
+	c.UpdateVolStatus()
 }
--- a/controller/control.go
+++ b/controller/control.go
@@ -709,17 +709,11 @@ getCloneStatus:
 	} else if status == "error" {
 		return fmt.Errorf("Replica clone status returned error %s", address)
 	}
-	c.setReplicaModeNoLock(address, types.WO)
 	go c.monitorPreload(address)
 	return nil
 }
 
 func (c *Controller) Start(addresses ...string) error {
-	var (
-		expectedRevision int64
-		sendSignal       int
-	)
-
 	c.Lock()
 	defer c.Unlock()
 
@@ -743,55 +737,29 @@ func (c *Controller) Start(addresses ...string) error {
 			return err
 		}
 	}
+	/*
+	   * This code has been commented since we start with only one replica currently
+	   	revisionCounters := make(map[string]int64)
+	   	for _, r := range c.replicas {
+	   		counter, err := c.backend.GetRevisionCounter(r.Address)
+	   		if err != nil {
+	   			logrus.Errorf("GetRevisionCounter failed %s %v", r.Address, err)
+	   			return err
+	   		}
+	   		if counter > expectedRevision {
+	   			expectedRevision = counter
+	   		}
+	   		revisionCounters[r.Address] = counter
+	   	}
 
-	revisionCounters := make(map[string]int64)
-	for _, r := range c.replicas {
-		counter, err := c.backend.GetRevisionCounter(r.Address)
-		if err != nil {
-			logrus.Errorf("GetRevisionCounter failed %s %v", r.Address, err)
-			return err
-		}
-		if counter > expectedRevision {
-			expectedRevision = counter
-		}
-		revisionCounters[r.Address] = counter
-	}
-
-	for address, counter := range revisionCounters {
-		if counter != expectedRevision {
-			logrus.Errorf("Revision conflict detected! Expect %v, got %v in replica %v. Mark as ERR",
-				expectedRevision, counter, address)
-			c.setReplicaModeNoLock(address, types.ERR)
-		}
-	}
-	for regrep := range c.RegisteredReplicas {
-		sendSignal = 1
-		for _, tmprep := range c.replicas {
-			if tmprep.Address == "tcp://"+regrep+":9502" {
-				sendSignal = 0
-				break
-			}
-		}
-		if sendSignal == 1 {
-			logrus.Infof("sending add signal to %v", regrep)
-			c.factory.SignalToAdd(regrep, "add")
-		}
-	}
-	for regrep := range c.RegisteredQuorumReplicas {
-		sendSignal = 1
-		for _, tmprep := range c.quorumReplicas {
-			if tmprep.Address == "tcp://"+regrep+":9502" {
-				sendSignal = 0
-				break
-			}
-		}
-		if sendSignal == 1 {
-			logrus.Infof("sending add signal to quorum %v", regrep)
-			c.factory.SignalToAdd(regrep, "add")
-		}
-	}
-	logrus.Info("Update volume status")
-	c.UpdateVolStatus()
+	   	for address, counter := range revisionCounters {
+	   		if counter != expectedRevision {
+	   			logrus.Errorf("Revision conflict detected! Expect %v, got %v in replica %v. Mark as ERR",
+	   				expectedRevision, counter, address)
+	   			c.setReplicaModeNoLock(address, types.ERR)
+	   		}
+	   	}
+	*/
 	return nil
 }
 
@@ -1039,15 +1007,67 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
 	c.RemoveReplicaNoLock(address)
 }
 
+// monitorPreload This function should only be called for the first replica
 func (c *Controller) monitorPreload(address string) {
-	c.Lock()
-	defer c.Unlock()
+	var (
+		sendSignal bool
+		status     string
+		err        error
+	)
 	logrus.Infof("Get preload status of backend %s", address)
-	if err := c.backend.GetPreloadStatus(address); err != nil {
-		logrus.Errorf("Failed to monitor preload, error: %v", err)
-		return
+	verifyPreload := true
+	for verifyPreload {
+		c.Lock()
+		if status, err = c.backend.GetPreloadStatus(address); err != nil {
+			logrus.Errorf("Failed to monitor preload, error: %v", err)
+			c.Unlock()
+			return
+		}
+		c.Unlock()
+		switch types.PreloadStatus(status) {
+		case types.Done:
+			verifyPreload = false
+			logrus.Infof("Preload is completed on backend %s", address)
+		case types.Started:
+			logrus.Warningf("Preload is not done yet on backend %s, current status: %s", address, status)
+			time.Sleep(2 * time.Second)
+		case types.Error:
+			logrus.Errorf("Preload status of backend is %s", status)
+			return
+		default:
+			logrus.Warningf("Preload is not done yet on backend %s, current status: %s", address, status)
+			time.Sleep(2 * time.Second)
+		}
 	}
+	c.Lock()
 	c.startFrontend()
 	c.setReplicaModeNoLock(address, types.RW)
 	c.UpdateVolStatus()
+	for regrep := range c.RegisteredReplicas {
+		sendSignal = true
+		for _, tmprep := range c.replicas {
+			if tmprep.Address == "tcp://"+regrep+":9502" {
+				sendSignal = false
+				break
+			}
+		}
+		if sendSignal {
+			logrus.Infof("sending add signal to %v", regrep)
+			c.factory.SignalToAdd(regrep, "add")
+		}
+	}
+	for regrep := range c.RegisteredQuorumReplicas {
+		sendSignal = true
+		for _, tmprep := range c.quorumReplicas {
+			if tmprep.Address == "tcp://"+regrep+":9502" {
+				sendSignal = false
+				break
+			}
+		}
+		if sendSignal {
+			logrus.Infof("sending add signal to quorum %v", regrep)
+			c.factory.SignalToAdd(regrep, "add")
+		}
+	}
+	c.Unlock()
 }
--- a/controller/control.go
+++ b/controller/control.go
@@ -1040,6 +1040,8 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
 }
 
 func (c *Controller) monitorPreload(address string) {
+	c.Lock()
+	defer c.Unlock()
 	logrus.Infof("Get preload status of backend %s", address)
 	if err := c.backend.GetPreloadStatus(address); err != nil {
 		logrus.Errorf("Failed to monitor preload, error: %v", err)
--- a/controller/control.go
+++ b/controller/control.go
@@ -599,6 +599,9 @@ func (c *Controller) SetReplicaMode(address string, mode types.Mode) error {
 	case types.RW:
 		c.Lock()
 		defer c.Unlock()
+	case types.WO:
+		c.Lock()
+		defer c.Unlock()
 	default:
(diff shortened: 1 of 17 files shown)


utkarshmani1997:
@vishnuitta Generating extents in travis in very short time is not possible. I have already added a test case in litmus to verify this. https://github.com/openebs/litmus/pull/384

vishnuitta:
> @vishnuitta Generating extents in travis in very short time is not possible. I have already added a test case in litmus to verify this. [openebs/litmus#384](https://github.com/openebs/litmus/pull/384)

lets add automated tests to hit our modified code.. not need to get lot of extents

sync/sync.go
```diff
@@ -358,8 +368,10 @@ Register:
 	}
 
 	logrus.Infof("reloadAndVerify %v", replicaAddress)
-	return t.reloadAndVerify(replicaAddress, toClient)
-
+	if err := t.reloadAndVerify(replicaAddress, toClient); err != nil {
+		return err
+	}
+	return s.Preload()
```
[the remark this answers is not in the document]
utkarshmani1997: Done

sync/sync.go
```diff
@@ -358,8 +362,24 @@ Register:
...
+		return err
+	}
 
+	t.preload(s)
+	return nil
+}
+
+func (t *Task) preload(s *replica.Server) {
+	logrus.Info("Start reading extents")
+	go func() {
```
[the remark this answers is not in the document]
vishnuitta: good that with controller restart testcase, we found this issue of monitoring disconnection.
Regarding closing of go-routine, lock will just safeguard not to enter, but, better would be reuse of same go routine without creating new one (or) to kill the go-routine.

sync/sync.go
```diff
@@ -358,8 +362,28 @@ Register:
...
+}
+
+func (t *Task) preload(s *replica.Server) {
+	logrus.Info("Start reading extents")
+	if s.PreloadStatus == "Started" {
+		logrus.Warning("Already reading extents in background")
+		return
+	}
+	go func() {
+		if err := s.Preload(); err != nil {
```
vishnuitta: have a testcase for error in preload

sync/sync.go
```diff
@@ -358,8 +362,28 @@ Register:
...
+		return err
+	}
 
+	t.preload(s)
+	return nil
+}
+
+func (t *Task) preload(s *replica.Server) {
+	logrus.Info("Start reading extents")
+	if s.PreloadStatus == "Started" {
```
vishnuitta: looks good.. will review it keenly

replica/backup.go
```diff
@@ -270,8 +272,12 @@ func preload(d *diffDisk) error {
 					length++
 				}
 			}
-			d.location[offset] = byte(i)
-			d.UsedBlocks++
+			d.rmLock.Lock()
+			if i != len(d.files) {
```
vishnuitta: shouldn't this be `d.location[offset] != len(d.files)`? and also, this assumes that new file is created after restart by the time preload starts. We need to handle that assumption.

replica/diff_disk.go
```diff
@@ -19,6 +19,7 @@ type diffDisk struct {
 	// mapping of sector to index in the files array. a value of 0 is special meaning
 	// we don't know the location yet.
 	location          []byte
+	preloadStatus     string
```
vishnuitta: make this as enum

vishnuitta: also, make it visible in some rest api

replica/diff_disk.go
```diff
@@ -311,7 +312,7 @@ func (d *diffDisk) lookup(sector int64) (byte, error) {
 	}
 
 	target := d.location[sector]
-	if target == 0 {
+	if target == 0 || d.preloadStatus == "Started" || d.preloadStatus == "Error" {
```
vishnuitta: better would be d.prelodStatus != "Done"

replica/diff_disk.go
vishnuitta: also comment explaining each state

replica/server.go
```diff
@@ -102,6 +104,20 @@ func (s *Server) Create(size int64) error {
...
+	if s.r == nil {
+		return fmt.Errorf("Can't read extents, s.r is nil")
+	}
+
+	s.PreloadStatus = "Started"
+	inject.AddPreloadTimeout()
+	if err := s.r.Preload(); err != nil {
+		return err
+	}
+	s.PreloadStatus = ""
```
vishnuitta: why do we need to nullify this status?

replica/server.go
```diff
@@ -102,6 +104,20 @@ func (s *Server) Create(size int64) error {
 	return r.Close()
 }
 
+func (s *Server) Preload() error {
+	if s.r == nil {
+		return fmt.Errorf("Can't read extents, s.r is nil")
+	}
+
+	s.PreloadStatus = "Started"
```
vishnuitta: To avoid multiple of these go routines to get started, PreloadStatus need to be set to 'Started' even before this go routine is created. In case if you want to know that go-routine is started, you can have some other state.

sync/sync.go
vishnuitta: As mentioned in one of above comment, PreloadStatus need to be set to 'Started' right after this check and before go routine is started.

replica/backup.go
utkarshmani1997: I was ignoring the last file completely, although checking at particular offset is better.

replica/diff_disk.go
utkarshmani1997: done

replica/diff_disk.go
utkarshmani1997: ok

sync/sync.go
utkarshmani1997: Sounds good

replica/backup.go
```diff
@@ -270,8 +275,12 @@ func preload(d *diffDisk) error {
 					length++
 				}
 			}
-			d.location[offset] = byte(i)
-			d.UsedBlocks++
+			d.rmLock.Lock()
+			if d.location[offset] < byte(len(d.files)) {
```
vishnuitta: is this supposed to be d.location[offset] < (byte(len(d.files)) -1 )

utkarshmani1997: d.location[offset] != (byte(len(d.files))) is correct validation, explanation is as follows:
- d.files[0] = <nil>
- d.files[1]  = file1 
and so on
on each restart a new file get appended to d.files so the latest file will be at the last location. For example if there was 5 files initially and restart happens so total file is 5+1(new file) = 6 and size of d.files would be 6 + 1(<nil> oth location) = 7. So we will verify that if offset is in latest file or not, if it's already  there then we will ignore and if it's not there will update it.

replica/diff_disk.go
```diff
@@ -242,6 +242,8 @@ func (d *diffDisk) fullReadAt(buf []byte, offset int64) (int, error) {
 		c   int
 	)
 
+	d.rmLock.Lock()
```
vishnuitta: this can be very long lock. Why do we need this?

sync/sync.go
```diff
@@ -381,8 +381,7 @@ func (t *Task) preload(s *replica.Server) {
 	go func() {
 		if err := s.Preload(); err != nil {
 			s.PreloadStatus = types.Error
-			logrus.Error(err)
-			s.MonitorChannel <- struct{}{}
+			logrus.Fatal("Exiting, err: ", err)
```
vishnuitta: have more explanatory log here

replica/server.go
```diff
@@ -376,6 +395,18 @@ func (s *Server) Close(signalMonitor bool) error {
 		return nil
 	}
 
+	if s.PreloadStatus != types.Done {
```
vishnuitta: why do we need a 'if' and then 'for' when you are however doing a 'log' in both the cases?

replica/backup.go
```diff
@@ -270,8 +275,12 @@ func preload(d *diffDisk) error {
 					length++
 				}
 			}
-			d.location[offset] = byte(i)
-			d.UsedBlocks++
+			d.rmLock.Lock()
+			if d.location[offset] < byte(i) {
+				d.location[offset] = byte(i)
+				d.UsedBlocks++
```
vishnuitta: is UsedBlocks coming correct now even with this change?

sync/sync.go
utkarshmani1997: err itself will be self explainatory.

replica/server.go
utkarshmani1997: I had used it for debugging purpose, removing it.

replica/backup.go
utkarshmani1997: We are not using `usedblock` in ci. `du -sh` is used to verify the sizes.

vishnuitta: ok.. so, to fix this still, we can do like
```
if d.location[offset] != byte(i) {
        d.UsedBlocks++
}
```
And, is there a way to know the value of this d.UsedBlocks during runtime?

replica/diff_disk.go
```diff
@@ -186,7 +183,9 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 				length++
 			}
 		}
+		d.rmLock.Lock()
```
vishnuitta: `UsedBlocks` seems to be going wrong due to this and previous 'space reclaimation' PR.
Can you explain me the usage of `d.UsedBlocks`?

replica/diff_disk.go
```diff
@@ -166,7 +163,7 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 		if val := d.location[startSector+i]; val == 0 {
 			d.UsedLogicalBlocks++
 			d.UsedBlocks++
-		} else if val != target {
+		} else if val != target && d.preloadStatus == types.Done {
```
vishnuitta: you need to add a comment here, and also, at the end of the 'for' loop where sendToCreateHole is called

replica/backup.go
vishnuitta: it seems - writeAt is also incrementing this..
so, do `d.UsedBlocks++` here, and we can do similar above check there during writes

utkarshmani1997:
Revert to earlier approach, i.e, replica will be WO mode until Preload get completed. IO's will be served if quorum is met. (2 replica in RW mode out of 3)

controller/replicator.go
```diff
@@ -505,3 +506,27 @@ func (r *replicator) GetCloneStatus(address string) (string, error) {
...
 	return status, nil
 }
+
+func (r *replicator) GetPreloadStatus(address string) error {
+	backend, ok := r.backends[address]
+	if !ok {
+		return fmt.Errorf("Cannot find backend %v", address)
+	}
+	for true {
+		status, err := backend.backend.GetPreloadStatus()
```
vishnuitta: can we avoid using `backend` here?

replica/server.go
```diff
@@ -376,6 +392,16 @@ func (s *Server) Close(signalMonitor bool) error {
 		return nil
 	}
 
+	if s.r.volume.preloadStatus != types.Done {
+		logrus.Warning("Already reading extents in background")
+		// verify if goroutine exited safely
+		for s.r.volume.preloadStatus == types.Started {
```
vishnuitta: this `for` loop doesn't make sense

controller/control.go
```diff
@@ -1035,3 +1035,14 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
...
 }
+
+func (c *Controller) monitorPreload(address string) {
+	logrus.Infof("Get preload status of backend %s", address)
+	if err := c.backend.GetPreloadStatus(address); err != nil {
+		logrus.Errorf("Failed to monitor preload, error: %v", err)
+		return
+	}
+	c.startFrontend()
+	c.setReplicaModeNoLock(address, types.RW)
```
vishnuitta: why 'NoLock'?

controller/replicator.go
utkarshmani1997: We can't, this is the way, we can get the PreloadStatus.

replica/server.go
utkarshmani1997: Why it doesn't make sense ?

controller/control.go
utkarshmani1997: There is no specific reason, since setReplicaModeNoLock() was getting used that's why used the same, because logic is no different that earlier.

vishnuitta: its not about that current code. Won't we be needing lock here?

utkarshmani1997: We are entering by taking lock, that's why we donot need another lock.

controller/control.go
```diff
@@ -1035,3 +1038,14 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
 	logrus.Infof("Monitoring stopped %v", address)
 	c.RemoveReplicaNoLock(address)
 }
+
+func (c *Controller) monitorPreload(address string) {
```
payes: why are we not using any locks in this function.

controller/control.go
```diff
@@ -706,7 +709,8 @@ getCloneStatus:
 	} else if status == "error" {
 		return fmt.Errorf("Replica clone status returned error %s", address)
 	}
-	c.setReplicaModeNoLock(address, types.RW)
+	c.setReplicaModeNoLock(address, types.WO)
```
payes: I guess this is not required as the replica is already in WO mode.

utkarshmani1997: This is required because we update the status to RO in UpdateVolStatus

controller/control.go
utkarshmani1997: have added in the recent commit

sync/sync.go
```diff
@@ -357,9 +361,19 @@ Register:
...
+	if _, err := toClient.ReloadReplica(); err != nil {
+		logrus.Errorf("Error in reloadreplica %s", replicaAddress)
+		return err
+	}
+
+	// Running as a saparate goroutine such that if controller
+	// disconnected while preload, replica get notified immediately
+	// on MonitorChannel.
+	logrus.Info("Preload and verify rebuild")
+	go t.preloadAndVerifyRebuild(replicaAddress, toClient, s)
```
payes: What is the need to run it as a goroutine.

controller/control.go
```diff
@@ -1035,3 +1038,16 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
 	logrus.Infof("Monitoring stopped %v", address)
 	c.RemoveReplicaNoLock(address)
 }
+
+func (c *Controller) monitorPreload(address string) {
+	c.Lock()
```
payes: Start controller and 3 replicas, immediately restart all 3 replicas.
Between the above 2 events, controller sends start signal to replica 1, and it is waiting for replica 1 to complete its preload with the lock and sleeping for 2 sec.
Replicas got restarted during this 2 sec, and the old IPs persist.
Controller again tries to get preload status on same IP after 2 sec, since controller doesn't know that the replica went down while it was sleeping. 
Preload status received by controller "Preload is not done yet on backend tcp://172.18.0.5:9502, current status: NA" Error at replica for the the above preloadStatus request "Fail to get revision counter: BUG: revision file wasn't initialized".
Controller tries it indefinitely while the GET request from replicas to controller are waiting for this lock.

sync/sync.go
```diff
@@ -322,7 +322,11 @@ Register:
 	}
 	if action == "start" {
 		logrus.Infof("Received start from controller")
-		return t.client.Start(replicaAddress)
+		if err := t.client.Start(replicaAddress); err != nil {
+			return err
+		}
+		go s.Preload()
```
payes: Care also needs to be taken while calling s.Preload() without a lock. If s.r is closed while preload is going on, replica might show unexpected behavior.

sync/sync.go
```diff
@@ -383,22 +397,25 @@ func (t *Task) checkAndResetFailedRebuild(address string, server *replica.Server
 	return nil
 }
 
-func (t *Task) reloadAndVerify(address string, repClient *replicaClient.ReplicaClient) error {
-	_, err := repClient.ReloadReplica()
-	if err != nil {
-		logrus.Errorf("Error in reloadreplica %s", address)
-		return err
+func (t *Task) preloadAndVerifyRebuild(address string, repClient *replicaClient.ReplicaClient, s *replica.Server) {
+	if err := s.Preload(); err != nil {
```
payes: Same as above comment

controller/control.go
```diff
@@ -741,56 +737,29 @@ func (c *Controller) Start(addresses ...string) error {
...
-		if sendSignal == 1 {
-			logrus.Infof("sending add signal to quorum %v", regrep)
-			c.factory.SignalToAdd(regrep, "add")
-		}
-	}
-	logrus.Info("Update volume status")
-	c.UpdateVolStatus()
-
+	/*
+	   * This code has been commented since we start with only one replica currently
```
vishnuitta: This commented code can be removed.. and, add a comment that this is moved to monitorPreload function

controller/control.go
```diff
@@ -1035,3 +1006,68 @@ func (c *Controller) monitoring(address string, backend types.Backend) {
...
+		sendSignal = true
+		for _, tmprep := range c.replicas {
+			if tmprep.Address == "tcp://"+regrep+":9502" {
+				sendSignal = false
+				break
+			}
+		}
+		if sendSignal {
+			logrus.Infof("sending add signal to %v", regrep)
+			c.factory.SignalToAdd(regrep, "add")
```
vishnuitta: This signaling after preload is important. In previous versions, this was missed, and would have caused some data issue.

types/types.go
```diff
@@ -134,6 +140,7 @@ type Frontend interface {
 }
 
 type DataProcessor interface {
+	Close(signalMonitor bool) error
```
vishnuitta: I don't think this change is required now.

replica/replica.go
```diff
@@ -173,7 +173,7 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 		r.info.BackingFileName = backingFile.Name
 	}
 	r.volume.sectorSize = defaultSectorSize
-
+	r.volume.preloadStatus = types.None
```
vishnuitta: lets add a log at all the status change of preload to help in debugging

replica/replica.go
```diff
@@ -217,11 +217,20 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	r.insertBackingFile()
 	r.ReplicaType = replicaType
 
+	return r, r.writeVolumeMetaData(true, r.info.Rebuilding)
```
vishnuitta: need to look at other callers of construct (clone? revert? etc) so that Preload is calling after this call

replica/server.go
```diff
@@ -86,6 +88,9 @@ func (s *Server) Create(size int64) error {
 	state, _ := s.Status()
 
 	if state != Initial {
+		r := &Replica{}
```
vishnuitta: why this change?

utkarshmani1997:
Closing this PR, since it doesn't cover clone and revert cases. Raised new PR #181 for the same. This PR can be opened if need arise in future.

PR #175, opened 2019-01-24
[TA4759] fix(replica): handle case of disk-detach

utkarshmani1997:
- Issue Ref: openebs/openebs#1387
- Log and handle the case of disk-detach cases (input/output error)
  in case it's `EIO` error fatal the replica.

Note:
- What happens to running replica ?
  Replica will be fataled.
- What will happen upon replica restart after fatal?
  Replica will always be fataled until disk is attached
- What will happen to controller ?
  It will remove the fataled replica, and will continue to serve
  IO's if quorum is met.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -3,6 +3,8 @@ package rpc
 import (
 	"io"
 	"net"
+	"os"
+	"syscall"
 	"time"
 
 	"github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


utkarshmani1997:
@vishnuitta following is the error, i got in GKE
```
time="2019-01-24T11:36:17Z" level=error msg="Failed to write data, error: write /openebs/volume-head-001.img: input/output error" 
time="2019-01-24T11:36:17Z" level=fatal msg=Exiting... 
```

vishnuitta:
can you cherry-pick this to v0.8.x branch as well?

vishnuitta:
@utkarshmani1997 this is needed for v0.8.x. Can you please cherry-pick and close issue#1387?

PR #179, opened 2019-01-28
fix(CI): Change the reference from $REPO to $BASELINE_REPO.

yudaykiran:
- The push scripts of Jiva already use the $REPO variables, which gets
overwritten by gitlabs $REPO value causing the build to fail.
- Renamed the $REPO variable in gitlab to $BASELINE_REPO in gitlab.

Signed-off-by: Uday Kiran Y <uday.kiran@openebs.io>

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,14 +26,14 @@ baseline-image:
      - echo $BRANCH
      - export COMMIT=${CI_COMMIT_SHA:0:7}
      - echo $COMMIT
-     - git clone https://github.com/$REPO/e2e-infrastructure.git
+     - git clone https://github.com/$BASELINE_REPO/e2e-infrastructure.git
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -26,14 +26,14 @@ baseline-image:
      - echo $BRANCH
      - export COMMIT=${CI_COMMIT_SHA:0:7}
      - echo $COMMIT
-     - git clone https://github.com/$REPO/e2e-infrastructure.git
+     - git clone https://github.com/$BASELINE_REPO/e2e-infrastructure.git
```
utkarshmani1997: @yudaykiran can we put a comment over this about where the value of $BASELINE_REPO will be initialized ?

yudaykiran: @utkarshmani1997 Its mentioned in the commit message, its internal to gitlab.

PR #181, opened 2019-02-12
[TA4197] fix(replica): preload issue due to large no of extents

utkarshmani1997:
Recently we have seen that if volume size is large then preload
takes time and hence openReplica fails due to timeouts and hence
neither of them succedes.

There are five calls where Preload is required

* When replica is in Initial state
* Replica received `open` signal from controller
* Replica is closed
* Replica is reloading
* Revert to snapshot

To solve the issue, we have removed http client timeout, and fixed
other issues which was depending on this.

Feature : Added `PreloadStatus` field in Replica structure, so status can be queried via REST api
There are three status:

```
Error    : In case of replica disconnection or syscall failure
Started  : If Preload is happening in the background
Done     : If Preload is completed.
```

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -98,9 +98,13 @@ func AutoConfigureReplica(s *replica.Server, frontendIP string, address string, 
 			continue
 		}
 		if err = AutoRmReplica(frontendIP, address); err != nil {
+			logrus.Warning("AutoRmReplica failed, retry after 5 second")
+			time.Sleep(5 * time.Second)
 			continue
 		}
 		if err = AutoAddReplica(s, frontendIP, address, replicaType); err != nil {
+			logrus.Warning("AutoAddReplica failed, retry after 5 second")
+			time.Sleep(5 * time.Second)
 			continue
 		}
 
--- a/app/replica.go
+++ b/app/replica.go
@@ -8,7 +8,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
-	"sync"
 	"syscall"
 	"time"
 
@@ -85,9 +84,12 @@ func CheckReplicaState(frontendIP string, replicaIP string) (string, error) {
 }
 
 func AutoConfigureReplica(s *replica.Server, frontendIP string, address string, replicaType string) {
+	var (
+		err   error
+		state string
+	)
 	for {
-		wg := sync.WaitGroup{}
-		state, err := CheckReplicaState(frontendIP, address)
+		state, err = CheckReplicaState(frontendIP, address)
 		logrus.Infof("Replicastate: %v err:%v", state, err)
 		if err == nil && (state == "" || state == "ERR") {
 			s.Close(false)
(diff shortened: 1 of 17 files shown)


app/replica.go
```diff
@@ -84,29 +84,46 @@ func CheckReplicaState(frontendIP string, replicaIP string) (string, error) {
...
+	for {
+		state, err = CheckReplicaState(frontendIP, address)
+		logrus.Infof("Replicastate: %v err:%v", state, err)
+
+		if err != nil {
+			logrus.Infof("checkReplicaState failed, err:%v retry after 5sec..", err)
+			time.Sleep(5 * time.Second)
+			continue
+		}
+		if state == "WO" || state == "ERR" {
```
vishnuitta: we want to autormreplica for WO also?
Please add comments if we need to this, and also, add a log that autormreplica is being attempted

app/replica.go
```diff
@@ -84,29 +84,46 @@ func CheckReplicaState(frontendIP string, replicaIP string) (string, error) {
...
-	logrus.Infof("Waiting on MonitorChannel")
-	select {
-	case <-s.MonitorChannel:
+		if err = AutoAddReplica(s, frontendIP, address, replicaType); err != nil {
+			logrus.Warning("AutoAddReplica failed, retry after 5 second")
+			time.Sleep(5 * time.Second)
+			continue
+		}
+
+		logrus.Infof("Waiting on MonitorChannel")
```
vishnuitta: We are waiting on monitorchannel so that RPC will write onto this? or any other go routine does this?

app/replica.go
vishnuitta: we want to autormreplica for WO also?
Please add comments if we need to this, and also, add a log that autormreplica is being attempted

backend/remote/remote.go
```diff
@@ -249,14 +256,16 @@ func (rf *Factory) Create(address string) (types.Backend, error) {
...
 	}
 
-	rpc := rpc.NewClient(conn, r.closeChan)
-	r.IOs = rpc
+	remote := rpc.NewClient(conn, r.closeChan)
+	r.IOs = remote
 
 	if err := r.open(); err != nil {
+		remote.SetRWExitTrue()
+		remote.Close()
```
vishnuitta: if open is not success, should we do close?

vishnuitta: got it.. both are different

rpc/client.go
```diff
@@ -23,6 +23,10 @@ var (
 	opUnmapTimeout  = 30 * time.Second // client unmap
 	opPingTimeout   = 20 * time.Second
 	opUpdateTimeout = 15 * time.Second // client update
+	// upon testing we have observed that replica was
+	// taking 1 sec for reading one thousand extents
+	// per file.
+	pingDeadline = 600 * time.Second // client update
```
vishnuitta: are we using this?

replica/replica.go
```diff
@@ -1155,6 +1160,12 @@ func (r *Replica) GetCloneStatus() string {
 	return info.CloneStatus
 }
 
+func (r *Replica) GetPreloadStatus() types.PreloadStatus {
```
vishnuitta: Do we need this?

replica/rest/model.go
```diff
@@ -25,6 +27,7 @@ type Replica struct {
 	UsedLogicalBlocks string                      `json:"usedlogicalblocks"`
 	UsedBlocks        string                      `json:"usedblocks"`
 	CloneStatus       string                      `json:"clonestatus"`
+	PreloadStatus     types.PreloadStatus         `json:"preloadStatus"`
```
vishnuitta: Do we need this?

app/replica.go
utkarshmani1997: If a replica is attached and sync file failed due to unreachable source replica this will help, i will add logs and comments for the same.

app/replica.go
utkarshmani1997: Correct, rpc and s.Close both can write on it.

backend/remote/remote.go
utkarshmani1997: correct, we need to close rpc connection in that case.

rpc/client.go
utkarshmani1997: No, thanks for pointing this out, will remove that.

replica/replica.go
utkarshmani1997: yes, adding this in replica info. Will be good for observability purpose.

replica/rest/model.go
utkarshmani1997: No, i will remove it.

utkarshmani1997:
Closing this as #183 will take care of this. A new PR will be raised for the other fixes.

PR #185, opened 2019-02-17
fix(replica): modify autoregistration process

utkarshmani1997:
1. Handle error for AutoAddReplica and AutoRmReplica
2. Fix logic for autoregistration in replica

Corner cases which this PR covers when replica goes for autoregistration : 
1. When this is first replica to go for autoregistration after restart:
    - open may take sometime and meanwhile controller get disconnected, rpc connection is closed and hence notified on monitorChannel, so replica will again check for it's state but will get empty state and hence will close itself
    - above case is valid for other replicas but it again proceeds to do some additional steps, so there are following edge cases where replica can stuck and until unless those get completed, replica wouldn't be proceed for autoregistration:
        * Create Replica : post a request to controller to be opened, so if controller get disconnected that request will get i/o timeout and replica would go for autoregistration.
        *  Sync : until unless sync get completed replica wouldn't go for autoregistration since there is no communication with controller in this process.
        *  ReloadAndVerify: reload is very similar to open call and replica get stuck here also

Fixes openebs/openebs#2399

### DIFF
--- a/app/add_replica.go
+++ b/app/add_replica.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"errors"
-	"time"
 
 	"github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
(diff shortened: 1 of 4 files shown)


utkarshmani1997:
closing via #216

PR #195, opened 2019-04-02
fix(controller): add missing flag StartSignalled in case of error

utkarshmani1997:
Fixes: openebs/openebs#2482

- Replica is not able to register again if signalled to start already
but startSignal() failed. This is due to missing flag StartSignalled
which was set to true.

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -272,15 +272,15 @@ func (c *Controller) registerReplica(register types.RegReplica) error {
 
 	if c.StartSignalled == true {
 		if c.MaxRevReplica == register.Address {
-			logrus.Infof("Replica %v signalled to start again, registered replicas: %v", c.MaxRevReplica, c.RegisteredReplicas)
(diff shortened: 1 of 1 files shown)


controller/control.go
```diff
@@ -269,12 +269,11 @@ func (c *Controller) registerReplica(register types.RegReplica) error {
...
+
 	if c.StartSignalled == true {
 		if c.MaxRevReplica == register.Address {
-			logrus.Infof("Replica %v signalled to start again %d:%d", c.MaxRevReplica,
-				len(c.RegisteredReplicas), c.replicaCount)
+			logrus.Infof("Replica %v signalled to start again, registered replicas: %v", c.MaxRevReplica, c.RegisteredReplicas)
 			if err := c.signalReplica(); err != nil {
-				c.StartSignalled = false
 				return err
 			}
```
vishnuitta: is the bug here is - to return nil when signalReplica gets success?

utkarshmani1997: No @vishnuitta, it was there when we verify required no of replicas to start, there we were not setting StartSignalled to false in case of error.

PR #196, opened 2019-04-04
feat(ci): add test case to verify fix of PR#195

utkarshmani1997:
This commit add a script which verifies that controller doesn't stuck
while replica registration due to the missing flag c.StartSignalled = false in error case.
This case is very hard to reproduce, here is the sequence to reproduce it:
- Start controller and two replica
- Kill the replica which gets start signal
- Start the killed replica and inject delay in registration at replica side so that registration of replica happens twice (goto statement get executed)
- Once you get two start signals from replica, kill it.
- You will be able to see following in the logs `Can signal only to  ,can't signal to`

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -49,6 +49,7 @@ func (r *Remote) Close() error {
 
 func (r *Remote) open() error {
 	logrus.Infof("Opening: %s", r.Name)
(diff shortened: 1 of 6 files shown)


utkarshmani1997:
Closing this PR, updated the PR description with the steps to reproduce.

PR #202, opened 2019-04-23
fix(replica): bad file descriptor error

utkarshmani1997:
Recently it has been observed that replicas were crashing due to
frequent controller restart. This is happening due to the files
are getting closed just after the disconnection but it is trying to
punch holes on the closed files. So it was giving `bad file descriptor
error`.

This PR fix the above issue by closing the files gracefully after
draining the queued data in the HoleCreatorChan. Punch holes will
be done only in following case:
- If replica is master replica (which get start signal)
- If replica is reloaded
- If replica is in RW mode (i.e, Rebuilding is false)

Additionally it fixes some go vet errors also.

Fixes: openebs/openebs#2532

### DIFF
--- a/ci/start_init_test.sh
+++ b/ci/start_init_test.sh
@@ -659,40 +659,49 @@ test_replica_rpc_close() {
 }
 
 verify_bad_file_descriptor_error() {
-        echo "-----------------Verify_bad_file_descriptor_error-------------"
-        verify_replica_cnt "2" "Two replica count test"
-        if [ "$1" == "When punching holes while preload" ];
-        then
-            run_ios_rand_write
-            docker stop $orig_controller_id
-            sleep 1
-            docker start $orig_controller_id
-            sleep 25
-        else
-            run_ios 200k 4k
-            docker stop $orig_controller_id
-            sleep 1
-            docker start $orig_controller_id
+	echo "-----------------Verify_bad_file_descriptor_error-------------"
+	verify_replica_cnt "2" "Two replica count test"
+	# Generate sufficient no of extents so that punch hole get
+	# delayed and we can verify that data which is filled in the
+	# HoleCreatorChan during preload is drained while closing the replica.
+	# Replica will be closed upon controller disconnection
+	run_ios_rand_write "1"
+	docker stop $orig_controller_id
+	sleep 1
+	docker start $orig_controller_id
+	verify_replica_cnt "2" "Two replica count test when controller is stopped in test_bad_file_descriptor"
+	verify_vol_status "RW" "When there are 2 replicas and controller is stopped in test_bad_file_descriptor"
 
-	    verify_replica_cnt "2" "Two replica count test when controller is stopped in test_bad_file_descriptor"
-	    verify_vol_status "RW" "When there are 2 replicas and controller is stopped in test_bad_file_descriptor"
-            run_ios 200k 4k
-            docker stop $orig_controller_id
-            sleep 1
-            docker start $orig_controller_id
-            sleep 30
-        fi
+
+	docker stop $replica1_id
+	replica3_id=$(start_replica "$CONTROLLER_IP" "$REPLICA_IP1" "vol1")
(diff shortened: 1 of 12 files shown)


replica/backup.go
```diff
@@ -222,18 +237,25 @@ func sendToCreateHole(f types.DiffDisk, offset int64, len int64) {
 	HoleCreatorChan <- hole
 }
 
-//Preload creates a mapping of block number to fileIndx (d.location).
-//This is done with the help of extent list fetched from filesystem.
-//Extents list in each file is traversed and the location table is updated
+func shouldCreateHoles() bool {
```
vishnuitta: add comments and why also?

replica/backup.go
```diff
@@ -188,6 +188,16 @@ type Hole struct {
 
 var HoleCreatorChan = make(chan Hole, 1000000)
 
+func drainHoleCreatorChan() {
```
vishnuitta: add comments

replica/server.go
```diff
@@ -376,6 +381,15 @@ func (s *Server) Close(signalMonitor bool) error {
 		return nil
 	}
 
+	types.DrainOps = types.DrainStart
```
vishnuitta: comments please..

vishnuitta:
@utkarshmani1997 BCH is failing. Can you look at it?

ci/start_init_test.sh
```diff
@@ -611,6 +623,60 @@ test_replica_rpc_close() {
...
 
+verify_bad_file_descriptor_error() {
+        echo "-----------------Verify_bad_file_descriptor_error-------------"
+        verify_replica_cnt "2" "Two replica count test"
+        if [ "$1" == "When punching holes while preload" ];
+        then
+            run_ios_rand_write
+            docker stop $orig_controller_id
+            sleep 1
+            docker start $orig_controller_id
```
vishnuitta: preload will not cause any impact.. we need to do IOs and restart controller one more time.. and also need to restart replica not in debug mode

utkarshmani1997: ok, sounds good

utkarshmani1997:
@vishnuitta BCH is failing since we are adding some variable inside a function which already has more than 30 lines of code. This can be fixed as a separate PR as a part of refactoring later.

sync/sync.go
```diff
@@ -322,13 +324,19 @@ Register:
 	}
 	if action == "start" {
 		logrus.Infof("Received start from controller")
-		return t.client.Start(replicaAddress)
+		types.IsMasterReplica = true
+		if err := t.client.Start(replicaAddress); err != nil {
```
vishnuitta: good catch

vishnuitta:
@payes can you review these changes?

replica/backup.go
```diff
@@ -222,18 +241,27 @@ func sendToCreateHole(f types.DiffDisk, offset int64, len int64) {
 	HoleCreatorChan <- hole
 }
 
-//Preload creates a mapping of block number to fileIndx (d.location).
-//This is done with the help of extent list fetched from filesystem.
-//Extents list in each file is traversed and the location table is updated
+// shouldCreateHoles returns bool which is used to as prerequisit for
+// punching holes.
+func shouldCreateHoles() bool {
```
payes: Lets add a TODO here to revisit this function in case Open Call is optimized in future, where the replica keeps itself open and waits for connection from controller.

sync/sync.go
```diff
@@ -322,13 +324,19 @@ Register:
...
+			return err
+		}
+		return nil
 	}
 	logrus.Infof("CheckAndResetFailedRebuild %v", replicaAddress)
 	if err := t.checkAndResetFailedRebuild(replicaAddress, s); err != nil {
 		return fmt.Errorf("CheckAndResetFailedRebuild failed, error: %s", err.Error())
 	}
 
+	types.IsRebuilding = true // Used to ignore punch hole
```
payes: Lets reset this global variable in case of errors in this function.

PR #204, opened 2019-04-24
feat(replica): optimize resync

utkarshmani1997:

This commit optimize the resync by checking the revision count. If the revision count of both the replicas
are same and the chain of both healthy and restarted replica is same, it wouldn't go for sync, rather it would be directly added RW replica after verifying the rebuild.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -355,7 +355,7 @@ Register:
 	}
 	inject.PanicAfterPrepareRebuild()
 
-	ok, err := t.isRevisionCountAndChainSame(replicaAddress, fromClient, toClient)
+	ok, err := t.isRevisionCountAndChainSame(fromClient, toClient)
 	if err != nil {
 		return err
 	}
@@ -372,7 +372,7 @@ Register:
 
 }
 
-func (t *Task) isRevisionCountAndChainSame(addr string, fromClient, toClient *replicaClient.ReplicaClient) (bool, error) {
+func (t *Task) isRevisionCountAndChainSame(fromClient, toClient *replicaClient.ReplicaClient) (bool, error) {
 	rwReplica, err := fromClient.GetReplica()
 	if err != nil {
 		return false, err
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -343,21 +343,9 @@ Register:
 		return fmt.Errorf("failed to get transfer clients, error: %s", err.Error())
 	}
 
-	logrus.Infof("SetRebuilding %v", replicaAddress)
+	logrus.Infof("SetRebuilding to true in %v", replicaAddress)
 	if err := toClient.SetRebuilding(true); err != nil {
 		return fmt.Errorf("failed to set rebuilding: true, error: %s", err.Error())
-	}
-
-	ok, err := t.isRevisionCountAndChainSame(replicaAddress, fromClient, toClient)
-	if err != nil {
-		return err
-	}
-
-	if ok {
-		if err := toClient.SetRebuilding(false); err != nil {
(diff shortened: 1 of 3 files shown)


sync/sync.go
```diff
@@ -346,6 +347,18 @@ Register:
 		return fmt.Errorf("failed to set rebuilding: true, error: %s", err.Error())
 	}
 
+	ok, err := t.isRevisionCountAndChainSame(replicaAddress, fromClient, toClient)
+	if err != nil {
+		return err
+	}
+
+	if ok {
```
vishnuitta: Due to this, do we miss out punching holes during preload, as there won't be reload?

sync/sync.go
```diff
@@ -346,6 +347,18 @@ Register:
...
+		if err := toClient.SetRebuilding(false); err != nil {
+			return fmt.Errorf("Failed to set rebuilding: true, error: %v", err)
+		}
+		return nil
+	}
+
 	logrus.Infof("PrepareRebuild %v", replicaAddress)
 	output, err := t.client.PrepareRebuild(rest.EncodeID(replicaAddress))
```
vishnuitta: how about doing PrepareRebuild as well, as it does sync of head meta files?

sync/sync.go
```diff
@@ -362,6 +375,31 @@ Register:
...
+		return false, err
+	}
+
+	if rwReplica.RevisionCounter == curReplica.RevisionCounter {
+		if !reflect.DeepEqual(rwReplica.Chain[1:], curReplica.Chain[1:]) {
+			logrus.Errorf("Replica %v's chain not equal to RW replica %v's chain", toClient.GetAddress(), fromClient.GetAddress())
+			return false, nil
+		}
+
+		if err := t.client.VerifyRebuildReplica(rest.EncodeID(addr)); err != nil {
```
vishnuitta: doing this is required as controller as well need to change the state of replica in its structures..
good one :+1:

sync/sync.go
```diff
@@ -362,6 +375,31 @@ Register:
...
+		return false, err
+	}
+
+	curReplica, err := toClient.GetReplica()
+	if err != nil {
+		return false, err
+	}
+
+	if rwReplica.RevisionCounter == curReplica.RevisionCounter {
+		if !reflect.DeepEqual(rwReplica.Chain[1:], curReplica.Chain[1:]) {
```
vishnuitta: why are we not considering Chain[0:]?

vishnuitta:
@payes review please..

sync/sync.go
utkarshmani1997: Chain[0] is head file

sync/sync.go
utkarshmani1997: @vishnuitta yes, you are right, we may miss few holes. How about calling reloadAndVerify here ?

sync/sync.go
utkarshmani1997: No @vishnuitta , PrepareRebuild only returns the chain[0:] but we compare content of the chain whether it starts with prefix `volume-head`, we ignore the sync of head files (both head and meta).

sync/sync.go
```diff
@@ -352,16 +353,45 @@ Register:
...
+			return err
+		}
+	}
+
 	logrus.Infof("reloadAndVerify %v", replicaAddress)
 	return t.reloadAndVerify(replicaAddress, toClient)
 
 }
 
+func (t *Task) isRevisionCountAndChainSame(addr string, fromClient, toClient *replicaClient.ReplicaClient) (bool, error) {
```
golangcibot: `(*Task).isRevisionCountAndChainSame` - `addr` is unused (from `unparam`)

PR #205, opened 2019-04-25
feat(logs): remove logs from stats endpoint

utkarshmani1997:
This commit removes the unnecessary logs generated for getting stats via rest calls.

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -182,8 +182,9 @@ func startReplica(c *cli.Context) error {
 		server := rest.NewServer(s)
 		router := http.Handler(rest.NewRouter(server))
 		router = util.FilteredLoggingHandler(map[string]struct{}{
-			"/ping":          {},
-			"/v1/replicas/1": {},
+			"/ping":                   {},
+			"/v1/replicas/1":          {},
+			"/v1/replicas/1/volusage": {},
 		}, os.Stdout, router)
 		logrus.Infof("Listening on control %s", controlAddress)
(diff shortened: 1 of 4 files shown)


app/replica.go
```diff
@@ -182,8 +182,9 @@ func startReplica(c *cli.Context) error {
 		server := rest.NewServer(s)
 		router := http.Handler(rest.NewRouter(server))
 		router = util.FilteredLoggingHandler(map[string]struct{}{
-			"/ping":          {},
-			"/v1/replicas/1": {},
+			"/ping":                   {},
+			"/v1/replicas/1":          {},
+			"/v1/replicas/1/volusage": {},
```
vishnuitta: why are we adding to filteredPaths?

replica/rest/replica.go
```diff
@@ -53,7 +53,6 @@ func (s *Server) GetReplicaStats(apiContext *api.ApiContext) *types.Stats {
 }
 
 func (s *Server) GetUsage(apiContext *api.ApiContext) (*types.VolUsage, error) {
-	logrus.Infof("GetUsage")
```
vishnuitta: remove log from
`ListReplicas`
`GetReplica`
`GetReplicaStats`
`GetUsage`
`GetStats`
`GetVolUsage`

utkarshmani1997: ok

app/replica.go
utkarshmani1997: So that logs from those paths can't be collected.

PR #216, opened 2019-05-14
fix(replica): remove MonitorChan and improve autoregistration logic

utkarshmani1997:
This commit fixes the issue, where replica is getting blocked on
MonitorChan for forever while registration/sync process if controller
is restarted or removed the replicas due to network disconnection
or any other explicit reason.

Now replicas will be fataled as soon as controller is disconnected
after closing the files gracefully.

Upgrade scenarios have been changed also, now user should first upgrade
the controller, which will stop the replicas and then upgrade the
replicas one by one.

### DIFF
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -1,7 +1,6 @@
 package rpc
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -15,10 +14,6 @@ import (
 )
 
 type operation func(*Message)
-
-var (
-	errInvalidReq = errors.New("Received invalid req")
-)
 
 type Server struct {
 	wire        *Wire
@@ -57,9 +52,6 @@ func (s *Server) Handle() error {
 		select {
 		case err = <-ret:
 			if err != nil {
-				if err == errInvalidReq {
-					logrus.Warningf("Failed to serve client: %v, rejecting request", s.wire.conn.RemoteAddr())
-				}
 				return err
 			}
 		case <-ticker.C:
@@ -122,9 +114,10 @@ func (s *Server) readWrite(ret chan<- error) {
 			// This is done to ignore the requests from invalid clients
 			// such as Prometheus which by default scraps from all the open
 			// ports.
-			if msg.MagicVersion != MagicVersion {
-				ret <- errInvalidReq
-				break
+			if msg != nil {
+				if msg.MagicVersion != MagicVersion {
+					logrus.Warningf("Failed to serve client: %v, rejecting request, invalid client", s.wire.conn.RemoteAddr())
+				}
 			}
 			ret <- err
 			break
--- a/rpc/server.go
+++ b/rpc/server.go
@@ -58,11 +58,7 @@ func (s *Server) Handle() error {
 		case err = <-ret:
 			if err != nil {
 				if err == errInvalidReq {
-					logrus.Warningf("Failed to serve client: %v, rejecting request, err: %v", s.wire.conn.RemoteAddr(), err)
-					// ignore error, connection may be closed from the other side.
-					_ = s.Stop() // close connection with the invalid client
-					// return nil since, no need to close files
-					return nil
+					logrus.Warningf("Failed to serve client: %v, rejecting request", s.wire.conn.RemoteAddr())
 				}
 				return err
 			}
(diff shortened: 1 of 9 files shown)


sync/sync.go
```diff
@@ -300,6 +300,10 @@ func (t *Task) AddReplica(replicaAddress string, s *replica.Server) error {
 		return fmt.Errorf("failed to create temp server, error: %s", err.Error())
 	}
 Register:
+	// replica might be in WO state and opened files but failed to
+	// get added to controller since rebuild is happening on other replicas
+	// please read the comments below for more info.
+	s.Close()
```
golangcibot: Error return value of `s.Close` is not checked (from `errcheck`)

sync/sync.go
```diff
@@ -341,7 +345,15 @@ Register:
...
-		return err
+		logrus.Errorf("Failed to create replica, error: %s", err.Error())
+		// cases for above failure:
+		// - controller is not reachable
+		// - replica is already added (remove replica in progress)
+		// - error while creating snapshot, to start fresh
+		// - replica might be in errored state
+		// - other replica might be in WO mode, and
+		// we can only have one WO replica at a time
+		goto Register
```
payes: what is the reason for this change ? this failure also leading to fatal is looking better.

utkarshmani1997: Reason is mentioned in the comments, crashing replica is not good in case, controller is not adding it because it can only add one "WO" replica at a time. If we are good with crash for this case then i can remove this.

payes: Then better put a sleep before retrying, otherwise it will max out the CPU

utkarshmani1997: ok

utkarshmani1997: ok

rpc/server.go
```diff
@@ -68,12 +72,9 @@ func (s *Server) Handle() error {
...
 			}
-			if time.Since(s.pingRecvd) >= opPingTimeout*2 {
-				// Close the connection as Ping is not recieved
-				// since long time after replica is opened.
-				if err := s.Stop(); err != nil {
-					return err
-				}
+			since := time.Since(s.pingRecvd)
+			if since >= opPingTimeout*2 {
+				return fmt.Errorf("Couldn't received ping since: %v", since)
```
payes: Didn't receive ping since

rpc/server.go
```diff
@@ -120,6 +121,18 @@ func (s *Server) readWrite(ret chan<- error) {
 			break
 		} else if err != nil {
 			logrus.Errorf("Failed to read: %v", err)
+			// check if client is valid, since MagicVersion will always
+			// be a constant for valid clients such as jiva controller.
+			// This is done to ignore the requests from invalid clients
+			// such as Prometheus which by default scraps from all the open
+			// ports.
+			msg, errRead := s.wire.ReadMessage()
```
payes: I don't understand the need for this code, can you please explain a bit more over here. There already is a check in s.wire.Read() which verifies the same.

utkarshmani1997: yes, but there it will return the error and fatal the replica, which we don't want to do if someone is making invalid requests on data connection.

utkarshmani1997: That's why we are verifying whether request is having valid header i.e, msg.version (although this is also not a correct approach, since anyone can send the same version in the payload while making request, but it is still better approach IMO, otherwise we need to TLS server to validate the clients, which we want to avoid at this point of time)

payes: You have already read MagicVersion from the wire in func s.wire.Read() and now there is no version on wire to be read. Please correct me if I am wrong.

utkarshmani1997: @payes you're right. Have updated the PR

rpc/server.go
```diff
@@ -120,6 +121,15 @@ func (s *Server) readWrite(ret chan<- error) {
 			break
 		} else if err != nil {
 			logrus.Errorf("Failed to read: %v", err)
+			// check if client is valid, since MagicVersion will always
+			// be a constant for valid clients such as jiva controller.
+			// This is done to ignore the requests from invalid clients
+			// such as Prometheus which by default scraps from all the open
+			// ports.
+			if msg.MagicVersion != MagicVersion {
+				ret <- errInvalidReq
```
payes: Now there is no need for this, since we fatal after the first erred connection.

PR #222, opened 2019-06-08
fix(replica): optimize performace by avoiding defer in data path

utkarshmani1997:
Improve performace by removing defer in the data path.
Defer statements is an overhead and should be avoided if
some critical section is getting executed very frequently.

Ref:  https://medium.com/i0exception/runtime-overhead-of-using-defer-in-go-7140d5c40e32


utkarshmani1997:
closing this PR, will reopen if required

PR #223, opened 2019-06-13
fix(replica): failed to find meta file error

utkarshmani1997:
This commit fixes the issue where replicas were not getting
attached to controller due to "Failed to find metadata for volume-snap*
during open" error. This was happening due to broken sync operation.

fixes: openebs/openebs#2469

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -2,6 +2,8 @@ package sync
 
 import (
 	"fmt"
+	"os"
+	"path"
 	"reflect"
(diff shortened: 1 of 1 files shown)


utkarshmani1997:
The root cause of this issue is if more than one replica get scheduled on the same node, it will get into the same situation. To avoid this, we are using podAntiAffinity and this issue is not observed post addition of the antiAffinity feature.

PR #224, opened 2019-06-14
fix(jiva): handle corner cases for snapshot delete

utkarshmani1997:
This commit fixes the following corner cases around delete snapshot command
- return error if it's the latest snapshot
- Optimizing lun map, so that there won't be need of preload
- Drain the data for punch hole from HoleCreatorChan, since file will be
  deleted and closed, therefore avoiding case of bad file descriptor
  error.

Command to list snapshot: `longhorn snapshot ls`
Command to remove snapshot: `longhorn snapshot rm <snap_name>`

Note: We may miss punching holes for few blocks, if it's already in
progress and delete snapshot request comes up.

Test cases that need to be covered:
1. H->S2->S1->S0:
   - Where S0 is autogenerated, S1 is autogenerated but both have data
     on different offset. S2 is user created with same data as S0 and
     delete S1.
   - Where S0 is autogenerated, S1 is autogenerated but both have data
     on different offset. S2 is user created with same data as S0 and
     delete S0.
   - Where S0 is autogenerated, S1 is usercreated but have data on the
     same offsets. S2 is autogenerated and data on the same offset and
     delete S1.
    - Where S0 is autogenerated, S1 is usercreated but have data on the
     same offsets. S2 is autogenerated and data on the same offset and
     delete S0.

### DIFF
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -189,6 +189,17 @@ type Hole struct {
 
 var HoleCreatorChan = make(chan Hole, 1000000)
 
+func holeDrainer() {
+	types.DrainOps = types.DrainStart
+	HoleCreatorChan <- Hole{}
+	for {
+		if types.DrainOps == types.DrainDone {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+}
+
 // drainHoleCreatorChan is called if replica is closed, to drain
 // all the data for punching hole in the HoleCreatorChan.
 func drainHoleCreatorChan() {
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -226,6 +226,9 @@ func CreateHoles() error {
 			types.DrainOps = types.DrainDone
 			continue
 		}
+		if (Hole{}) == hole {
+			continue
+		}
 		fd = hole.f.Fd()
 	retry:
 		if err := syscall.Fallocate(int(fd),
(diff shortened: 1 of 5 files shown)


replica/replica.go
```diff
@@ -354,11 +361,15 @@ func (r *Replica) findDisk(name string) int {
 func (r *Replica) RemoveDiffDisk(name string) error {
 	r.Lock()
 	defer r.Unlock()
-
+	r.holeDrainer()
```
vishnuitta: add the comments

replica/backup.go
```diff
@@ -189,6 +189,17 @@ type Hole struct {
 
 var HoleCreatorChan = make(chan Hole, 1000000)
 
+func holeDrainer() {
```
vishnuitta: duplicate code.. this can be called from Close() of replica/server.go

replica/backup.go
```diff
@@ -189,6 +189,17 @@ type Hole struct {
...
+	types.DrainOps = types.DrainStart
+	HoleCreatorChan <- Hole{}
```
vishnuitta: lets handle empty structure in CreateHoles()

utkarshmani1997: @vishnuitta not required, since we are setting types.DrainOps = types.DrainDone draining it completely, so no empty structure will be there.

replica/backup.go
utkarshmani1997: ok

PR #228, opened 2019-06-20
Too many snapshots

utkarshmani1997:
This commit increase the maximumChainLength default value and also adds capability 
to configure it via environment variable.

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -149,7 +149,7 @@ func startReplica(c *cli.Context) error {
 
 	types.MaxChainLength, _ = strconv.Atoi(os.Getenv("MAX_CHAIN_LENGTH"))
 	if types.MaxChainLength == 0 {
-		logrus.Infof("MAX_CHAIN_LENGTH env not set, default value is 250")
+		logrus.Infof("MAX_CHAIN_LENGTH env not set, default value is 512")
 	} else {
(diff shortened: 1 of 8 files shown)


app/replica.go
```diff
@@ -144,6 +147,13 @@ func startReplica(c *cli.Context) error {
 		return errors.New("directory name is required")
 	}
 
+	types.MaxChainLength, _ = strconv.Atoi(os.Getenv("MAX_CHAIN_LENGTH"))
+	if types.MaxChainLength == 0 {
+		logrus.Infof("MAX_CHAIN_LENGTH env not set, default value is 250")
```
kmova: 255?

utkarshmani1997: it's 250 @kmova

replica/replica.go
```diff
@@ -33,7 +33,7 @@ const (
 	diskPrefix         = "volume-snap-"
 	diskSuffix         = ".img"
 	diskName           = diskPrefix + "%s" + diskSuffix
-	maximumChainLength = 250
+	maximumChainLength = 300
```
kmova: Can we make this default as 512?

vishnuitta:
@utkarshmani1997 please cherry-pick this to v1.0.x and v0.9.x as well

Issue #230, opened 2019-06-25
Broken filesystem after batch `jivactl snapshot rm`


rgembalik:
I wanted to flatten jiva pvc as much as possible, as it consisted mostly out of empty snapshots. I entered pvc controller pod and executed `jivactl snapshot rm $(jivactl snapshot ls | tail -1)` a bunch of times in a loop, one after another. Initially everything looked good, but after some time mysql pod crashed (gralog logs attached). 

Once I detected it I deleted the pod and rescaled it again, and automatic fsck fixed fs errors, but that ended with half of data missing. Recovery from sql backup fixed the issue.
[mywql-graylog.csv.txt](https://github.com/openebs/jiva/files/3325369/mywql-graylog.csv.txt)
[mywql-graylog-pvc-ctrl.csv.txt](https://github.com/openebs/jiva/files/3328836/mywql-graylog-pvc-ctrl.csv.txt)
[mywql-graylog-pvc-replica-kube2.csv.txt](https://github.com/openebs/jiva/files/3328837/mywql-graylog-pvc-replica-kube2.csv.txt)
[mywql-graylog-pvc-replica-kube3.csv.txt](https://github.com/openebs/jiva/files/3328838/mywql-graylog-pvc-replica-kube3.csv.txt)

PR #231, opened 2019-07-01
fix(replica): fix broken fs due to delete snapshot

utkarshmani1997:
Fix: openebs/jiva#230

This commit fixes the issue of broken file system at the
application side.

This is happening due to the hardlinked files were pointing to the 
older fd's which is not valid anymore after the hardlinking of child
to parent snapshot, since it's still pointing to the older data.
So files need to be reopened and their reference should be updated
with the new one with the coalesced data.

### DIFF
--- a/ci/start_init_test.sh
+++ b/ci/start_init_test.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -x
+#set -x
 PS4='${LINENO}: '
 CONTROLLER_IP="172.18.0.2"
 REPLICA_IP1="172.18.0.3"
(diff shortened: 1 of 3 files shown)


replica/replica.go
```diff
@@ -426,6 +430,25 @@ func (r *Replica) ReplaceDisk(target, source string) error {
 		return err
 	}
 
+	// find the updated index of target
+	index := r.findDisk(target)
+	if index <= 0 {
+		return nil
```
vishnuitta: is this error, and should we do fatal?

replica/replica.go
```diff
@@ -426,6 +430,25 @@ func (r *Replica) ReplaceDisk(target, source string) error {
...
+	}
+
+	// Close the removed file
+	if err := r.volume.files[index].Close(); err != nil {
+		return err
```
vishnuitta: this and next errors also can be handled with fatal?

ci/start_init_test.sh
```diff
@@ -1735,8 +1735,153 @@ test_duplicate_data_delete() {
...
+        device_name=""
+        file1="f1"
+        file2="f2"
+        file3="f3"
+        hash_file1=""
+        hash_file2=""
+        hash_file3=""
+        hash_file1=$(run_dd "$file1" "0")
+        hash_file2=$(run_dd "$file2" "100000")
+	copy_files_into_mnt_dir "$file1" "$file2" &
```
vishnuitta: lets add sleep 5 after this for test to run

vishnuitta: sleep 2

PR #242, opened 2019-09-19
update(jiva): update vendor and use go modules 

utkarshmani1997:
- Update vendor with the latest stable version
- Remove trash and dapper
- Use go modules for dependency management
- Remove backup code since it's not getting used


.gitignore
```diff
@@ -9,3 +9,7 @@ integration/.idea/
 integration/.tox/
 integration/.venv/
 integration/MANIFEST
+cscope.files
+cscope.out
+f1
+f2
```
prateekpandey14: is it required ?

cscope.files
```diff
@@ -0,0 +1,693 @@
+/home/golearn/src/github.com/openebs/jiva/Makefile
+/home/golearn/src/github.com/openebs/jiva/app/replica.go
```
prateekpandey14: remove this file

libiscsi-testsuite/tags
```diff
@@ -0,0 +1,6 @@
+!_TAG_FILE_FORMAT	2	/extended format; --format=1 will not append ;" to lines/
+!_TAG_FILE_SORTED	1	/0=unsorted, 1=sorted, 2=foldcase/
```
prateekpandey14: we can remove this one as well

.gitignore
utkarshmani1997: yes, just to ignore the files used while integration testing and for other purpose

cscope.files
utkarshmani1997: ok

libiscsi-testsuite/tags
utkarshmani1997: ok

vishnuitta:
cc: @AmitKumarDas can you please review makefile/go modules?

PR #243, opened 2019-09-26
Alertlogs for Jiva Replica operations create/delete/snapshot/clone

Moteesh:
Signed-off-by: moteesh <moteesh@gmail.com>

PR #250, opened 2019-10-13
Refactor startReplica(c *cli.Context)

agarwalrounak:
Fixes https://github.com/openebs/openebs/issues/2774

Signed-off-by: Rounak Agarwal <rounakr.ag73@gmail.com>


app/replica.go
```diff
@@ -137,14 +137,71 @@ func CloneReplica(s *replica.Server, address string, cloneIP string, snapName st
...
+		logrus.Fatalf("Rest API exited: %v", resp)
+	case resp = <-rpcResp:
+		logrus.Fatalf("RPC listen exited: %v", resp)
+	case resp = <-syncResp:
+		logrus.Fatalf("Sync process exited: %v", resp)
+	}
+	return resp
+}
+
+func fetchValues(c *cli.Context) (string, string, *replica.Server, string, string, string, string, string, error) {
```
utkarshmani1997: returning too many values is not a good approach IMO, so create a structure and initialize the values and then return it to the callee.

utkarshmani1997: rename it to getStartConfig()

app/replica.go
```diff
@@ -164,6 +221,10 @@ func startReplica(c *cli.Context) error {
 	cloneIP := c.String("cloneIP")
 	snapName := c.String("snapName")
 	size := c.String("size")
+	return dir, replicaType, s, address, frontendIP, cloneIP, snapName, size, err
+}
+
+func handleSize(s *replica.Server, size string) error {
```
utkarshmani1997: rename it createReplica()

app/replica.go
```diff
@@ -177,7 +238,10 @@ func startReplica(c *cli.Context) error {
 			return err
 		}
 	}
+	return nil
+}
 
+func handleAddress(address string) (string, string, string, error) {
```
utkarshmani1997: same comment as above, try to use a structure for such use cases.

app/replica.go
```diff
@@ -246,55 +309,32 @@ func startReplica(c *cli.Context) error {
...
-					logrus.Errorf("Error in setting the clone status as 'error', found error:%v", statusErr)
-					return err
-				}
-				return err
-			}
-		}
-		logrus.Infof("Set clone status as Completed")
-		if err := s.Replica().SetCloneStatus("completed"); err != nil {
-			logrus.Error("Error in setting the clone status as 'completed'")
+func startClone(s *replica.Server, snapName, address, cloneIP string) error {
```
utkarshmani1997: good one

utkarshmani1997: can we pass the arguments as a structure `cloneConfig`

app/replica.go
agarwalrounak: Should the structure contain all the return values 
OR 
 only the string and *replica.Server values, and error is returned separately?

utkarshmani1997: error should be returned separately, all those string fields and *replica.Server will be part of a new structure

agarwalrounak: Got it. Thanks for the clarification.

app/replica.go
agarwalrounak: Instead of creating a new structure, can we utilize the structure(_to be created, not part of this code yet_) returned by the function `fetchValues`/`getStartConfig`? This returned structure already has the arguments being passed in the function `startClone` and a few more arguments.

agarwalrounak:
The Travis CI is failing for commit https://github.com/openebs/jiva/pull/243/commits/7fac4ee96fe379034c0affeffd8de3f3a420a943 where alertlogs were introduced in the code. It's the same reason why the Travis CI is failing here too.

utkarshmani1997:
@agarwalrounak `test -z "$(go fmt ${PACKAGES} | tee /dev/stderr)"` is failing for your case. It was passed in the commit you have mentioned. Can you try to fix it ?
You can run `go fmt ./app` to verify the same.

agarwalrounak:
@utkarshmani1997 When I ran `go fmt ./app`, the command returned `app/replica.go` along with the changes made in commit https://github.com/openebs/jiva/commit/9e3a446cbbb5958732e3571eec86f7ae6fe7ab22. After this change, running `go fmt ./app` now returns nothing. 
Can you check if the test you mentioned above is passing now?

app/replica.go
```diff
@@ -157,14 +166,96 @@ func CloneReplica(s *replica.Server, address string, cloneIP string, snapName st
 }
 
 func startReplica(c *cli.Context) error {
+	config, err := getStartConfig(c)
+	if err != nil {
+		return err
+	}
+	dir, replicaType, s, address, frontendIP, snapName, size := config.dir, config.replicaType, config.s,
+		config.address, config.frontendIP, config.snapName, config.size
```
utkarshmani1997: no need to create local variables, just use it where required

app/replica.go
```diff
@@ -157,14 +166,96 @@ func CloneReplica(s *replica.Server, address string, cloneIP string, snapName st
...
+	err = createReplica(s, size)
+	if err != nil {
+		return err
+	}
+
+	add, err := handleAddress(address)
+	if err != nil {
+		return err
+	}
+	controlAddress, dataAddress, syncAddress := add.controlAddress, add.dataAddress, add.syncAddress
```
utkarshmani1997: same comment as above

PR #244, opened 2019-09-30
Add build for arm

Wangzihao18:
Related to:
- openebs/openebs#1295

This PR modify the Makefile to determime the arch and os.
And add Dockerfile_build_arm64 and Dockerfile_build_amd64
to support build for arm64 arch.

Signed-off-by: wangzihao <wangzihao18@huawei.com>

PR #252, opened 2019-10-15
feat(build): Add build tags env for arm

Wangzihao18:
Signed-off-by: wangzihao <wangzihao18@huawei.com>

### DIFF
--- a/scripts/build_binaries
+++ b/scripts/build_binaries
@@ -6,4 +6,9 @@ source $(dirname $0)/version
 cd $(dirname $0)/..
 
 mkdir -p bin
-go build  -ldflags "-X main.VERSION=$VERSION" -o bin/longhorn
+if [ ${ARCH} == "linux_arm64" ]
+then
+    GOARCH=arm GOOS=linux GOARM=5 go build -ldflags "-X main.VERSION=$VERSION" -o bin/longhorn
+else
+    go build  -ldflags "-X main.VERSION=$VERSION" -o bin/longhorn  
+fi
--- a/scripts/build_debug_binaries
+++ b/scripts/build_debug_binaries
@@ -6,4 +6,9 @@ source $(dirname $0)/version
 cd $(dirname $0)/..
 
 mkdir -p bin/debug
-go build -tags "debug" -ldflags "-X main.VERSION=$VERSION" -o bin/debug/longhorn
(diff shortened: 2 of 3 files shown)


Wangzihao18:
Reference to https://github.com/openebs/jiva/pull/244#issuecomment-537812240
I also using gox to cross compiling https://github.com/Wangzihao18/jiva/commit/2594426f0229b4b5c5bac0f8d8c88df78d3907a3.
But it shows error when using arm to compile amd.
This change cannot build cross-compilation docker images. Maybe should it be support on the future features ?

utkarshmani1997:
@Wangzihao18 `ARCH` env is taken from the `go env GOARCH` in Makefile, so above change can not be used for the cross compilation, rather we can add targets in Makefile like `build_arm64` where the default value of `ARCH` can be changed to the `linux_arm64`. We don't build and push the images ourself, we just tag the repo with the appropriate tag and travis automatically build the binaries and push it to docker/quay etc.
So by adding a new target in Makefile, we can call the same in `.travis.yaml`. 
Note that we will not trigger `ci/start_init_test.sh` as this will require lot of other work like setting up travis of arm64 builds to test the same. We will add the test environment later in the future.

Regarding your other question related to building cross-compiled docker images, i need to learn how it works.

cc: @kmova @vishnuitta

Wangzihao18:
> @Wangzihao18 `ARCH` env is taken from the `go env GOARCH` in Makefile, so above change can not be used for the cross compilation, rather we can add targets in Makefile like `build_arm64` where the default value of `ARCH` can be changed to the `linux_arm64`. We don't build and push the images ourself, we just tag the repo with the appropriate tag and travis automatically build the binaries and push it to docker/quay etc.
> So by adding a new target in Makefile, we can call the same in `.travis.yaml`.
> Note that we will not trigger `ci/start_init_test.sh` as this will require lot of other work like setting up travis of arm64 builds to test the same. We will add the test environment later in the future.
> 
> Regarding your other question related to building cross-compiled docker images, i need to learn how it works.
> 
> cc: @kmova @vishnuitta

Ok, I got it. For now, it's better to download source code and build it. We have a Organization account `openestuary` in docker hub. Need I push my arm images to it ? or for your consideration, it's not ok that another organization push openebs images.

PR #254, opened 2019-10-16
Refactor scheduler extender-related API fields

Apollo9999:
For trace ability and historical reasons, it's vague that which scheduler APIs should be used internally, while which ones should be externalized, and which should be supported in a Kubernetes API's fashion, and which to serve as a plain protocol contract for HTTP communications only.

API cleanup is a line item , starting with this PR, we will address most of the API concerns in the scheduler.

Specifically for this PR, we're going to move all API fields related to extender implementation (not including Extender or Policy configuration APIs) to pkg/scheduler/apis/extender/v1


Makefile
```diff
@@ -87,7 +87,6 @@ ifndef golint
 	$(error "golint is not available in GOPATH. You can install it - go get -u golang.org/x/lint/golint ")
 endif
 	@echo "Linting with golint"
-	$(shell golint $(shell find . -maxdepth 1 -type d \( ! -iname ".git" \)) )
```
Apollo9999: $(shell golint $(shell find . -maxdepth 1 -type d \( ! -iname ".git" \)) )

vishnuitta:
Linting is required. Hence, marking this as Invalid, and closing.

PR #263, opened 2020-01-05
fix(replica): add replica with greater revision count

utkarshmani1997:
- This commit fixes an issue where replica was getting revision count
  value as zero since complete changes were not checked in into the last
  commit.
- Add ci to test the replica restart optimization

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -151,9 +151,7 @@ func (r *Remote) GetRevisionCounter() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	if replica.State != "open" && replica.State != "dirty" && replica.State != "rebuilding" {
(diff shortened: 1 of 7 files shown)


backend/remote/remote.go
```diff
@@ -151,7 +151,7 @@ func (r *Remote) GetRevisionCounter() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
-	if replica.State != "open" && replica.State != "dirty" {
+	if replica.State != "open" && replica.State != "dirty" && replica.State != "rebuilding" {
```
payes: Please state the necessity of the above and reverify all the scenarios where the replica goes down in rebuilding state.

utkarshmani1997: Thanks for pointing this out but this check is unnecessary now as we are persisting the revision count.

PR #265, opened 2020-01-13
feat(jiva): add options to enable buffered io in replica

utkarshmani1997:
This commit adds a flag in replicas and controller to enable buffered io. This improves the read/write performance significantly and recommended for the databases which send the sync periodically.
To enable the buffered io's one needs to pass following flag in the deployment of controller and replica.
`./jiva replica --buffered-io --frontendIP 127.0.0.1 --listen 127.0.0.1:9502 --size 2G vol1`
`./jiva controller --buffered-io --frontend gotgt --frontendIP 127.0.0.1 store1`

Limitations

- O_DIRECT does not guarantee that the data is written to the disk and the same is true for the alternative approach.
- Revision count will be increased immediately even though the data is not reached to the disk.
- Abrupt shutdown of the node may lead to partial data loss. It can be a few ms to seconds worth of the data (This scenario is not tested yet).
- Since we do PUNCH_HOLE on the files in case of overwrites in the background, replicas will be punching holes in the older snapshots even though the writes haven’t been successful in the latest file.
- May degrade the performance when heavy IO’s are being performed on the disk as hit ratio may go down so an additional overhead of querying io will be added to read from the page cache first and if not found then go back to the disk to read. Same is true for the write io’s as I/O wait might increase as well.

### DIFF
--- a/app/replica.go
+++ b/app/replica.go
@@ -2,7 +2,6 @@ package app
 
 import (
 	"errors"
-	"github.com/openebs/jiva/alertlog"
 	"net"
 	"net/http"
 	"os"
@@ -12,6 +11,8 @@ import (
 	"strings"
 	"syscall"
 	"time"
+
+	"github.com/openebs/jiva/alertlog"
 
 	"github.com/openebs/jiva/types"
 
@@ -53,6 +54,10 @@ func ReplicaCmd() cli.Command {
 			},
 			cli.BoolTFlag{
 				Name: "sync-agent",
+			},
+			cli.BoolFlag{
+				Name:  "buffered-io",
+				Usage: "if set true io's will be buffered, recommended for db's where data sync happens at intervals",
 			},
 			cli.StringFlag{
 				Name:  "size",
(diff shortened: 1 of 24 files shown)


app/replica.go
```diff
@@ -213,12 +210,74 @@ func startReplica(c *cli.Context) error {
...
+		return errors.New("directory name is required")
+	}
+
+	initLogger()
+	setMaxChainLength()
+
+	dir := c.Args()[0]
+	replicaType := c.String("type")
+	s := replica.NewServer(dir, 512, replicaType)
+	go replica.CreateHoles()
```
golangcibot: Error return value of `replica.CreateHoles` is not checked (from `errcheck`)

controller/control.go
```diff
@@ -163,10 +170,39 @@ func (c *Controller) hasWOReplica() (string, bool) {
...
 	var (
-		woRevCnt int64
-		err      error
+		woRevCnt, woSyncCnt int64
+		err                 error
 	)
 
+	if c.Bufio {
+		if _, ok := c.RegisteredReplicas[woReplica]; !ok {
+			woClient, err := replicaClient.NewReplicaClient(woReplica)
```
golangcibot: shadow: declaration of "err" shadows declaration at line 174 (from `govet`)

utkarshmani1997:
closing this PR as its not suitable for the environments where node crashes/ force shutdown is very frequent, so this may lead to unavailability of data, it could be from few ms to seconds. Please refer to following design [doc](https://docs.google.com/document/d/1qfPy3ZP62wiUiy0tKAwdAauHoxdwyfP6mlsiWV96cSk/edit?usp=sharing)

PR #268, opened 2020-01-29
[WIP] fix(replica): persist last IO offset and size

utkarshmani1997:
This commit fixes/improves following:
- WO replica was going for full rebuild in case if there is only one missing IO
  by persisting the offset and size of the last IO so that we can sync the last
  I/O during initializing the replica.

- Create autogenerated snapshots only when volume is in RW mode, since it
  doesnot make any sense to create a new snapshot even when volume is in RO,
  this was also making the resync of last missing IO bit complex, since new
  snapshots were created due to which the need of persisting file name was
  arising and again this name was getting changed after generating new snapshot.

- sync head file as well when volume is in RO mode

- Fixes recurring ReadOnly logs

Design Link: https://docs.google.com/document/d/1DzjIB2Fu_EvNZ4VpeWCUjaa37MEWZztbFXpaw9yYjUQ/edit

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -939,7 +939,7 @@ func (c *Controller) Start(addresses ...string) error {
 // on the app.
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	c.Lock()
-	if c.ReadOnly == true && !c.logged {
+	if c.ReadOnly && !c.logged {
 		c.logged = true
 		err := fmt.Errorf("Mode: ReadOnly")
 		c.Unlock()
--- a/controller/control.go
+++ b/controller/control.go
@@ -36,6 +36,7 @@ type Controller struct {
 	StartTime                time.Time
 	StartSignalled           bool
 	ReadOnly                 bool
+	logged                   bool
 	SnapshotName             string
 	IsSnapDeletionInProgress bool
 	StartAutoSnapDeletion    chan bool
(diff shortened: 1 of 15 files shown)


controller/control.go
```diff
@@ -935,12 +939,14 @@ func (c *Controller) Start(addresses ...string) error {
 // on the app.
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	c.Lock()
-	if c.ReadOnly == true {
+	if c.ReadOnly == true && !c.logged {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.ReadOnly` (from `gosimple`)

sync/sync.go
```diff
@@ -273,23 +274,98 @@ Register:
...
+	if volume.ReadOnly == "false" {
+		logrus.Infof("Volume is healthy, skip verifying last missing io")
+		return false, nil
+	}
+
+	if !isVersionCompatible(rwReplica.Version, volume.Version) {
+		return false, nil
+	}
+
+	var curRevCount int64 = 0
```
golangcibot: ineffectual assignment to `curRevCount` (from `ineffassign`)

utkarshmani1997:
Closing this PR due to following reasons:
- Upgrade related code needs to be added in data path
- It is not guaranteed that client will send the same IO if failed to write.
- Design constraints like syncing of head files if volume is in RO mode.
More details can be found out [here](https://docs.google.com/document/d/1DzjIB2Fu_EvNZ4VpeWCUjaa37MEWZztbFXpaw9yYjUQ/edit?usp=sharing)

PR #270, opened 2020-02-07
refact(jiva): restrict recurring ReadOnly logs by adding flag

utkarshmani1997:

### DIFF
--- a/controller/control.go
+++ b/controller/control.go
@@ -37,6 +37,7 @@ type Controller struct {
 	StartSignalled           bool
 	ReadOnly                 bool
 	SnapshotName             string
+	logged                   bool
 	IsSnapDeletionInProgress bool
 	StartAutoSnapDeletion    chan bool
 }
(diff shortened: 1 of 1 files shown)


controller/control.go
```diff
@@ -966,12 +969,14 @@ func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
...
 	c.Lock()
-	if c.ReadOnly == true {
+	if c.ReadOnly && !c.logged {
+		c.logged = true
 		err := fmt.Errorf("Mode: ReadOnly")
 		c.Unlock()
 		time.Sleep(1 * time.Second)
 		return -1, err
 	}
+	c.logged = false
```
payes: Let's also put a log saying, volume started serving IOs. Even though we have the log times when the volume became RO and RW, but we would also like to know the timestamps when the first time IOs were unserved and when the first time IOs were being served again.

payes: Let's also put a log saying, volume started serving IOs. Even though we have the log times when the volume became RO and RW, but we would also like to know the timestamps when the first time IOs were unserved and when the first time IOs were being served again.

controller/control.go
```diff
@@ -935,12 +936,14 @@ func (c *Controller) Start(addresses ...string) error {
 // on the app.
 func (c *Controller) WriteAt(b []byte, off int64) (int, error) {
 	c.Lock()
-	if c.ReadOnly == true {
+	if c.ReadOnly && !c.logged {
```
payes: We shouldn't skip this entire block, only logging needs to be disabled.

utkarshmani1997: good catch

utkarshmani1997:
closing this PR, as this is not solving the problem

PR #277, opened 2020-03-05
fix(build):arm images to be pushed to jiva-arm64

kmova:
Fixes an issue with https://github.com/openebs/jiva/pull/275

The images should be pushed to jiva-arm64. 

### DIFF
--- a/scripts/push
+++ b/scripts/push
@@ -8,10 +8,11 @@ then
 fi
 
 source "$(dirname $0)/version"
-if [ "$XC_ARCH" == "amd64" ]; then
-  IMAGEID=$( docker images -q openebs/jiva:${VERSION} )
-else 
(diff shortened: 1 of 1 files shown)


scripts/push
```diff
@@ -8,10 +8,11 @@ then
 fi
 
 source "$(dirname $0)/version"
-if [ "$XC_ARCH" == "amd64" ]; then
-  IMAGEID=$( docker images -q openebs/jiva:${VERSION} )
-else 
+if [ ${ARCH} == "linux_arm64" ]; then
```
shubham14bajpai: Can we have more generic condition for all other arch except amd64? 
`if [ ${ARCH} != "linux_amd64" ]`

kmova: I think it is better to fall back to the default (or current) behavior when it is not linux_arm64.

PR #278, opened 2020-03-06
feat(replica): run fsync on files & dir after create/remove/rename operation on files

utkarshmani1997:
This commit enhances the following:
-  Sync and close the replicas gracefully before fataling upon
error or killing the replicas by trapping SIGINT, SIGTERM signals
- Do fsync on files and directories after create/remove/rename operation

Note:
- Ref: https://github.com/etcd-io/etcd/issues/6368
- Trying to figure out the tools to simulate this behaviour but this is a must feature that we need to have.
- Implemented periodicSync goroutine to flush the metadata to the disk every 5 seconds, but disabled it for now due to performance costrains.


rpc/server.go
```diff
@@ -183,6 +183,7 @@ func (s *Server) handleRead(msg *Message) {
 	if err != nil {
 		logrus.Errorf("Failed to read data, error: %v", err)
 		if s.isIOError(err) {
+			s.data.Close()
```
golangcibot: Error return value of `s.data.Close` is not checked (from `errcheck`)

rpc/server.go
```diff
@@ -194,6 +195,7 @@ func (s *Server) handleWrite(msg *Message) {
 	if err != nil {
 		logrus.Errorf("Failed to write data, error: %v", err)
 		if s.isIOError(err) {
+			s.data.Close()
```
golangcibot: Error return value of `s.data.Close` is not checked (from `errcheck`)

rpc/server.go
```diff
@@ -211,6 +213,7 @@ func (s *Server) handleSync(msg *Message) {
 	if err != nil {
 		logrus.Errorf("Failed to sync data, error: %v", err)
 		if s.isIOError(err) {
+			s.data.Close()
```
golangcibot: Error return value of `s.data.Close` is not checked (from `errcheck`)

rpc/server.go
```diff
@@ -221,6 +224,7 @@ func (s *Server) handleUnmap(msg *Message) {
 	if err != nil {
 		logrus.Errorf("Failed to unmap data, error: %v", err)
 		if s.isIOError(err) {
+			s.data.Close()
```
golangcibot: Error return value of `s.data.Close` is not checked (from `errcheck`)

replica/backup.go
```diff
@@ -238,6 +238,7 @@ func CreateHoles() error {
 			time.Sleep(1)
 			retryCount++
 			if retryCount == 5 {
+				s.Close()
```
golangcibot: Error return value of `s.Close` is not checked (from `errcheck`)

replica/replica.go
```diff
@@ -457,6 +457,7 @@ func (r *Replica) ReplaceDisk(target, source string) error {
 	}
 
 	if err := r.rmDisk(source); err != nil {
+		r.close()
```
golangcibot: Error return value of `r.close` is not checked (from `errcheck`)

replica/replica.go
```diff
@@ -473,13 +474,15 @@ func (r *Replica) ReplaceDisk(target, source string) error {
 
 	// Close the removed file
 	if err := r.volume.files[index].Close(); err != nil {
+		r.close()
```
golangcibot: Error return value of `r.close` is not checked (from `errcheck`)

replica/replica.go
```diff
@@ -473,13 +474,15 @@ func (r *Replica) ReplaceDisk(target, source string) error {
...
 		logrus.Fatalf("Failed to close old instance of target: %v, err: %v", target, err)
 		return err
 	}
 
 	// Open for R/W
 	newFile, err := r.openFile(r.activeDiskData[index].Name, 0)
 	if err != nil {
+		r.close()
```
golangcibot: Error return value of `r.close` is not checked (from `errcheck`)

app/replica.go
```diff
@@ -178,7 +178,11 @@ func startReplica(c *cli.Context) error {
 	dir := c.Args()[0]
 	replicaType := c.String("type")
 	s := replica.NewServer(dir, 512, replicaType)
-	go replica.CreateHoles()
+	go replica.CreateHoles(s)
```
golangcibot: Error return value of `replica.CreateHoles` is not checked (from `errcheck`)

app/replica.go
```diff
@@ -178,7 +178,11 @@ func startReplica(c *cli.Context) error {
...
+	// close replica gracefully upon kill signal
+	addShutdown(func() {
+		s.Close()
```
golangcibot: Error return value of `s.Close` is not checked (from `errcheck`)

vishnuitta:
- Is this issue reproduced by performing node restarts and did we observed inconsistent blocks across replicas?
- If so, does this regular sync (say 5 seconds) helps (with the current way of checking for chain and meta files during sync process)?

replica/backup.go
```diff
@@ -238,6 +238,7 @@ func CreateHoles() error {
 			time.Sleep(1)
 			retryCount++
 			if retryCount == 5 {
+				_ = s.Close()
```
vishnuitta, line 241: this is NOT safe to do as other go routines still have fd and doing IOs on it

app/replica.go
```diff
@@ -178,7 +178,11 @@ func startReplica(c *cli.Context) error {
 	dir := c.Args()[0]
 	replicaType := c.String("type")
 	s := replica.NewServer(dir, 512, replicaType)
-	go replica.CreateHoles()
+	go replica.CreateHoles(s)
+	// close replica gracefully upon recieving SIGTERM or SIGINT
+	addShutdown(func() {
```
vishnuitta, line 183: when this func is called on signal, does all other go routines are exited (or) will they be still running?
If they will be still running, those go routines might be doing IOs on the fd available with them.

replica/backup.go
utkarshmani1997: This is safe to do @vishnuitta since s.Close will acquire lock and then close the fds. So I/O's will  get blocked and when lock is free those goroutine will check if s.r  will be set to nil.

app/replica.go
utkarshmani1997: This is handler function, so it will run s.Close upon SIGINT or SIGTERM signals. If any goroutine is already processing s.Close() will be blocked till those operations get finished and then files will be closed gracefully.

replica/replica.go
```diff
@@ -903,16 +960,16 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 
 	f, newHeadDisk, err := r.createNewHead(oldHead, newSnapName, created)
 	if err != nil {
+		r.doCleanup(newHeadDisk.Name)
```
vishnuitta, line 963: these changes are not related to this PR

replica/server.go
```diff
@@ -157,9 +159,9 @@ func (s *Server) Create(size int64) error {
 
 func (s *Server) Open() error {
 	s.Lock()
-	defer s.Unlock()
 
 	if s.r != nil {
+		s.Unlock()
```
vishnuitta, line 164: why these changes?

replica/replica.go
utkarshmani1997: This is to handle the corner case of cleaning up the leftovers due to error.

replica/server.go
utkarshmani1997: Ok, earlier there was go routine introduced so defer was not helpful. Will revert these changes.

replica/replica.go
```diff
@@ -736,19 +747,23 @@ func (r *Replica) encodeToFile(obj interface{}, file string) error {
...
 	defer f.Close()
 
 	if err := json.NewEncoder(f).Encode(&obj); err != nil {
 		logrus.Errorf("failed to encode the data to file: %s", f.Name())
 		return err
 	}
 
-	if err := f.Close(); err != nil {
-		logrus.Errorf("failed to close file after encoding to file: %s", f.Name())
+	if err := r.closeAndSyncDir(f); err != nil {
```
vishnuitta, line 758: one doubt: already 'f' is closed in closeAndSyncDir but defer does close again.. isn't this some issue?

replica/replica.go
```diff
@@ -909,7 +951,7 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 		if !done {
 			r.rmDisk(newHeadDisk.Name)
 			r.rmDisk(newSnapName)
-			f.Close()
+			f.Close() // rm only unlink the file since fd is still open
```
vishnuitta, line 954: better to do sync after linkDir also

replica/replica.go
```diff
@@ -783,6 +814,12 @@ func (r *Replica) createNewHead(oldHead, parent, created string) (types.DiffDisk
...
 		return nil, disk{}, err
 	}
+
+	err = r.syncDir()
+	if err != nil {
+		return nil, disk{}, fmt.Errorf("Fail to sync dir, err: %v", err)
+	}
+
 	if err := syscall.Truncate(r.diskPath(newHeadName), r.info.Size); err != nil {
 		return nil, disk{}, err
```
vishnuitta, line 824: add a comment that 'file just created may need to be deleted'

replica/replica.go
```diff
@@ -1115,10 +1157,13 @@ func (r *Replica) unmarshalFile(file string, obj interface{}) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
 
-	dec := json.NewDecoder(f)
-	return dec.Decode(obj)
+	err = json.NewDecoder(f).Decode(obj)
+	if err != nil {
+		return err
```
vishnuitta, line 1163: f.close is missing here

vishnuitta: why this change required?

utkarshmani1997: We were ignoring error thrown by f.Close().

replica/replica.go
```diff
@@ -1115,10 +1157,16 @@ func (r *Replica) unmarshalFile(file string, obj interface{}) error {
...
 		return err
 	}
-	defer f.Close()
 
-	dec := json.NewDecoder(f)
-	return dec.Decode(obj)
+	err = json.NewDecoder(f).Decode(obj)
+	if err != nil {
+		if closeErr := f.Close(); closeErr != nil {
+			logrus.Errorf("Fail to close file: %v, err: %v", f.Name(), err)
```
vishnuitta, line 1164: you may want to print closeErr instead of err

PR #280, opened 2020-03-09
revert(controller): comment out autosnapdeletion

utkarshmani1997:
- This commit comment out the autosnapshot deletion related code for now. This
  will be enabled by making it configurable in the upcoming releases.

- Fixes: openebs/openebs#2956


utkarshmani1997:
@payes let's do this in next release as if we miss something it will delay the release.

PR #281, opened 2020-03-09
revert(replica): comment revision count related checks at replica

utkarshmani1997:
- Comment out the revision count related checks as this is causing
  snapshot size mismatch issues where cluster instability is observed due to
  network or load on a node.
- comment out some basic tests as travis build time exceeds to 50 minutes.
- Fixes: openebs/openebs#2956

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -273,17 +273,17 @@ Register:
 	}
 	inject.PanicAfterPrepareRebuild()
 
-	ok, err := t.isRevisionCountAndChainSame(fromClient, toClient)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


sync/sync.go
```diff
@@ -273,17 +273,17 @@ Register:
 	}
 	inject.PanicAfterPrepareRebuild()
 
-	ok, err := t.isRevisionCountAndChainSame(fromClient, toClient)
-	if err != nil {
+	//	ok, err := t.isRevisionCountAndChainSame(fromClient, toClient)
```
vishnuitta: lets bring this change back

utkarshmani1997: ok

PR #287, opened 2020-03-24
fix(replica): get usedlogical size info from healthy replica

utkarshmani1997:
- This commit improves the logic to only get the usedlogical size from RW
  replica and additionally adds following info into stats
  * Chain info
  * UsedSize of snapshots
  * Detailed Snapshot info
- Fix the issue where revision count was showing as zero for the `jivactl
snapshot info` command.
- Move DiskInfo and ReplicaInfo to types.go as it is common across replica and controller
  for REST API calls.

### DIFF
--- a/controller/rest/volume.go
+++ b/controller/rest/volume.go
@@ -41,11 +41,16 @@ func (s *Server) GetVolume(rw http.ResponseWriter, req *http.Request) error {
 
 func (s *Server) GetReplicaInfo() []types.ReplicaInfo {
 	var (
-		info []types.ReplicaInfo
-		wg   sync.WaitGroup
+		info     []types.ReplicaInfo
+		replicas []types.Replica
+		wg       sync.WaitGroup
 	)
 	infoLock := &sync.Mutex{}
-	replicas := s.c.ListReplicas()
+	s.c.RLock()
+	for _, rep := range s.c.ListReplicas() {
+		replicas = append(replicas, rep)
+	}
+	s.c.RUnlock()
 	wg.Add(len(replicas))
 	for _, replica := range replicas {
 		addr := replica.Address
--- a/controller/rest/volume.go
+++ b/controller/rest/volume.go
@@ -4,9 +4,12 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/gorilla/mux"
+	replicaClient "github.com/openebs/jiva/replica/client"
+	"github.com/openebs/jiva/types"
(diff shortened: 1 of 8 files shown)


replica/rest/model.go
```diff
@@ -210,13 +196,16 @@ func NewReplica(context *api.ApiContext, state replica.State, info replica.Info,
 	r.SectorSize = info.SectorSize
 	r.Size = strconv.FormatInt(info.Size, 10)
 	r.RevisionCounter = strconv.FormatInt(info.RevisionCounter, 10)
+	r.UsedBlocks, r.UsedLogicalBlocks = "0", "0" // replica must be initializing
```
payes: Let's put NA over here instead of "0". "0" is misleading and shocking.
And let's mention in comment that this info  is still not available.

controller/rest/volume.go
```diff
@@ -36,6 +39,37 @@ func (s *Server) GetVolume(rw http.ResponseWriter, req *http.Request) error {
 	return nil
 }
 
+func (s *Server) GetReplicaInfo() []types.ReplicaInfo {
+	var (
+		info []types.ReplicaInfo
+		wg   sync.WaitGroup
+	)
+	infoLock := &sync.Mutex{}
+	replicas := s.c.ListReplicas()
```
payes: Need to take a lock and copy the list in a different variable.
Range might have unexpected behavior if a replica is removed from the list by some other function.

replica/rest/model.go
utkarshmani1997: It is converted to int64 later, so putting default value.

controller/rest/volume.go
utkarshmani1997: Good catch.

controller/rest/volume.go
```diff
@@ -36,6 +39,42 @@ func (s *Server) GetVolume(rw http.ResponseWriter, req *http.Request) error {
...
 
+func (s *Server) GetReplicaInfo() []types.ReplicaInfo {
+	var (
+		info     []types.ReplicaInfo
+		replicas []types.Replica
+		wg       sync.WaitGroup
+	)
+	infoLock := &sync.Mutex{}
+	s.c.RLock()
+	for _, rep := range s.c.ListReplicas() {
```
golangcibot: S1011: should replace loop with `replicas = append(replicas, s.c.ListReplicas()...)` (from `gosimple`)

replica/rest/model.go
payes: But make sure when user is fetching this value, he doesn't get 0.

PR #288, opened 2020-03-24
fix(replica): print http resp for bad status

utkarshmani1997:

### DIFF
--- a/backend/remote/remote.go
+++ b/backend/remote/remote.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


backend/remote/remote.go
```diff
@@ -113,7 +114,8 @@ func (r *Remote) doAction(action string, obj interface{}) error {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("Bad status: %d %s", resp.StatusCode, resp.Status)
+		content, _ := ioutil.ReadAll(resp.Body)
```
prateekpandey14: i get the idea of ignoring the error ,  but can we include the err check here , for example in case if there is err while reading the body, there will not be any read content anyways

utkarshmani1997: This will only show error if buffer is too large to hold, which is unlikely to happen since we don't have very large payload, that's why ignoring the error.

utkarshmani1997:
closing in favour  of #289

PR #290, opened 2020-04-01
feat(replica): add option to flush log to file

utkarshmani1997:
This commit add an option to enable persisting logs to a file. Following
options can be passed to configure log files:
- maxLogFileSize: It is the max size of log that can be hold by a file
  in megabites before rotation. Default is 100mb.
  If a write would cause the log file to be larger than MaxSize,
  the file is closed, renamed to include a timestamp of the current time,
  and a new log file is created using the original log file name.
  If the length of the write is greater than MaxSize, an error is returned.

- retentionPeriod: It is the max no of days to retain the old files.
  Default is 180 days.

- maxBackups: It is the maximum number of old log files to retain.
  Default is 5 log files.
- Added command to configure via `jivactl`

For example:
- `./jivactl replica --frontendIP 127.0.0.1 --listen :9502 --size 10g --logtofile=true --maxLogFileSize 1 --retentionPeriod 2 --maxBackups 4 vol1` (while starting replica)
- `jivactl logtofile enable --maxLogFileSize 1 --retentionPeriod 2 --maxBackups 4` on running volumes from controller 

Fixes: https://github.com/openebs/openebs/issues/2964

### DIFF
--- a/replica/server.go
+++ b/replica/server.go
@@ -131,10 +131,6 @@ func (s *Server) isExtentSupported() error {
 func (s *Server) Create(size int64) error {
 	s.Lock()
 	defer s.Unlock()
-	if err := os.Mkdir(s.dir, 0700); err != nil && !os.IsExist(err) {
-		logrus.Errorf("failed to create directory: %s", s.dir)
-		return err
-	}
 	if err := s.isExtentSupported(); err != nil {
 		return err
 	}
--- a/util/util.go
+++ b/util/util.go
@@ -82,10 +82,12 @@ func WriteLogInfo(dir string, lf LogToFile) error {
 		logrus.Errorf("failed to create temp file: %s while WriteLogInfo", LogInfo)
 		return err
 	}
-	defer f.Close()
 
 	if err := json.NewEncoder(f).Encode(&lf); err != nil {
 		logrus.Errorf("failed to encode the data to file: %s", f.Name())
+		if closeErr := f.Close(); closeErr != nil {
+			logrus.Errorf("failed to close file: %v, err: %v", f.Name(), closeErr)
+		}
 		return err
 	}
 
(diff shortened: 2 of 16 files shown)


kmova:
How are users going to use or interact with this feature? Editing deployment to enable or disable this feature is not an option for end-users.

replica/server.go
```diff
@@ -129,10 +129,6 @@ func (s *Server) isExtentSupported() error {
 func (s *Server) Create(size int64) error {
 	s.Lock()
 	defer s.Unlock()
-	if err := os.Mkdir(s.dir, 0700); err != nil && !os.IsExist(err) {
-		logrus.Errorf("failed to create directory: %s", s.dir)
-		return err
-	}
 	if err := s.isExtentSupported(); err != nil {
```
vishnuitta, line 132: for this `isExtentSuppoted`, does above 's.dir' need to be created?

vishnuitta: ok.. got to see that its created in the caller of 'Create' itself

vishnuitta:
> How are users going to use or interact with this feature? Editing deployment to enable or disable this feature is not an option for end-users.

@utkarshmani1997 can you update PR related to APIs in new changes?

utkarshmani1997:
> > How are users going to use or interact with this feature? Editing deployment to enable or disable this feature is not an option for end-users.
> 
> @utkarshmani1997 can you update PR related to APIs in new changes?

Added command to do the same, example is given in description.

util/util.go
```diff
@@ -20,12 +22,100 @@ import (
...
+	}
+
+	if !lf.Enable {
+		logrus.Infof("Disable logging to file")
+		return WriteLogInfo(dir, lf)
+	}
+	return StartLoggingToFile(dir, lf)
+}
+
+func LogRotate(dir string) error {
```
prateekpandey14, line 115: this function is not used any where ?

utkarshmani1997: Keeping it here for future usage, like if user wants to trigger logrotation without changing the existing policy.

PR #291, opened 2020-04-02
fix(replica): rm headfile if already exists

utkarshmani1997:
- This commit fix the issue where replica is not able to recover on
restart if new headfile created with the same name already exists.
- Add a test case to verify the same.

Fixes: https://github.com/openebs/openebs/issues/2907

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -813,7 +813,7 @@ func (r *Replica) createNewHead(oldHead, parent, created string) (types.DiffDisk
 			return nil, disk{}, fmt.Errorf("Can't remove head file %v as it contains some data", newHeadName)
 		}
 		if err = r.rmDisk(newHeadName); err != nil {
-			logrus.Errorf("Fail to remove disk, err: %v", err)
+			return nil, disk{}, fmt.Errorf("Failed to remove disk, err: %v", err)
 		}
 	}
 
@@ -846,10 +846,7 @@ func (r *Replica) linkDisk(oldname, newname string) error {
 
 	dest := r.diskPath(newname)
 	if _, err := os.Stat(dest); err == nil {
-		logrus.Infof("Old file %s exists, deleting", dest)
-		if err := os.Remove(dest); err != nil {
-			return err
-		}
+		return fmt.Errorf("Old file :%v already exists", newname)
 	}
 
 	if err := os.Link(r.diskPath(oldname), dest); err != nil {
@@ -883,16 +880,15 @@ func (r *Replica) rmDisk(name string) error {
 		return nil
 	}
 
-	lastErr := os.Remove(r.diskPath(name))
+	if err := os.Remove(r.diskPath(name)); err != nil {
+		return err
+	}
+
 	if err := os.Remove(r.diskPath(name + metadataSuffix)); err != nil {
-		lastErr = err
+		return err
 	}
 
-	if lastErr == nil {
-		lastErr = r.syncDir()
-	}
-
-	return lastErr
+	return r.syncDir()
 }
 
 func (r *Replica) revertDisk(parent, created string) (*Replica, error) {
@@ -918,7 +914,9 @@ func (r *Replica) revertDisk(parent, created string) (*Replica, error) {
 	}
 
 	// Need to execute before r.Reload() to update r.diskChildrenMap
-	r.rmDisk(oldHead)
+	if err := r.rmDisk(oldHead); err != nil {
+		return nil, err
+	}
 
 	rNew, err := r.Reload()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


replica/replica.go
```diff
@@ -776,7 +776,10 @@ func (r *Replica) createNewHead(oldHead, parent, created string) (types.DiffDisk
 	}
 
 	if _, err := os.Stat(r.diskPath(newHeadName)); err == nil {
```
payes, line 778: Can we also have a check for the space consumed by the file over here. The space consumed by the file should be zero. If not, we can sleep and fatal out.

utkarshmani1997: done

replica/replica.go
```diff
@@ -738,6 +738,8 @@ func (r *Replica) encodeToFile(obj interface{}, file string) error {
 	}
 	defer f.Close()
 
+	// crash at this point leads to stale .tmp file
```
vishnuitta, line 741: this comment may not be required as os.Create seems to be truncating the file if already exists..
can you please verify with a test program before adding any cleanup code here?

replica/replica.go
```diff
@@ -776,7 +778,13 @@ func (r *Replica) createNewHead(oldHead, parent, created string) (types.DiffDisk
 	}
 
 	if _, err := os.Stat(r.diskPath(newHeadName)); err == nil {
-		return nil, disk{}, fmt.Errorf("%s already exists", newHeadName)
+		logrus.Warningf("Head file: %v already exists", newHeadName)
+		if r.getDiskSize(newHeadName) > 0 {
+			return nil, disk{}, fmt.Errorf("Can't remove head file %v as it contains some data", newHeadName)
+		}
+		if err = r.rmDisk(newHeadName); err != nil {
+			logrus.Errorf("Fail to remove disk, err: %v", err)
```
vishnuitta, line 786: lets return error in case of rmDisk also which is supposed to delete the file successfully

vishnuitta: and change log message as: Failed to remove already existing disk

replica/replica.go
```diff
@@ -901,34 +909,32 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 		newSnapName = ""
 	}
 
+	// new head file created will be deleted in next call if replica crashes just after this
 	f, newHeadDisk, err := r.createNewHead(oldHead, newSnapName, created)
 	if err != nil {
+		rmDiskErr := r.rmDisk(newHeadDisk.Name)
```
vishnuitta, line 915: good one

replica/replica.go
```diff
@@ -776,7 +778,13 @@ func (r *Replica) createNewHead(oldHead, parent, created string) (types.DiffDisk
 	}
 
 	if _, err := os.Stat(r.diskPath(newHeadName)); err == nil {
-		return nil, disk{}, fmt.Errorf("%s already exists", newHeadName)
+		logrus.Warningf("Head file: %v already exists", newHeadName)
+		if r.getDiskSize(newHeadName) > 0 {
```
vishnuitta, line 782: good one..
and in linkDisk also, should we have similar check?

replica/replica.go
vishnuitta: In rmDisk, errNotExist can be handled as not an error

replica/replica.go
vishnuitta: in linkDisk, there should be some checks for metafile existence and delete them as well?

replica/replica.go
```diff
@@ -962,6 +969,23 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 	r.volume.UserCreatedSnap = append(r.volume.UserCreatedSnap, userCreated)
 	r.activeDiskData = append(r.activeDiskData, &newHeadDisk)
 
+	// assign it to temp variable, so that	r.info remains
+	// intact in case of crash
```
vishnuitta, line 973: in case of crash, everything in-memory is gone :).. so, not to worry here to use r.info

replica/replica.go
```diff
@@ -962,6 +969,23 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
...
+	info := r.info
+	info.Head = newHeadDisk.Name
+	info.Dirty = true
+	info.Parent = newSnapName
+	info.RevisionCounter = newHeadDisk.RevisionCounter
```
vishnuitta, line 978: is RevisionCounter missed in old code?

replica/replica.go
utkarshmani1997: @vishnuitta i tested here by panic call and after panic it executes all the functions in stack and close is being called which modifies the content of volume.meta. That's why added this.

replica/replica.go
utkarshmani1997: yes

replica/replica.go
utkarshmani1997: No @vishnuitta if file exists in linkdisk then that is a problem, file should not exist in the first place.

replica/replica.go
utkarshmani1997: @vishnuitta i have verified .tmp file will remain, if replica crashes at this point. @payes asked me to put this comment.

replica/replica.go
utkarshmani1997: No, because if metafile doesn't exist then it's a problem for all other cases like snapshot delete or snapshot create. In this case use case is different, because even though we were not able to remove a file, we are going to create a fresh file with the same name, returning an error doesn't make any sense as it could be a bad fd and could stuck here indefinitely on next iterations.

replica/replica.go
```diff
@@ -962,6 +969,23 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
...
+
+	// new head encoded to metadata
+	if err := r.encodeToFile(&info, volumeMetaData); err != nil {
+		return err
+	}
+
+	// crash at this point leads to stale old head and it meta file
```
payes, line 985: add a TODO over here
// Crash here, stale old head and corresponding metafile will remain,
// which will not be part of the chain. This needs to be removed on next
// restart

replica/replica.go
payes: Does this revision counter play any role ??

replica/replica.go
```diff
@@ -937,9 +943,11 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 		r.diskData[newSnapName].UserCreated = userCreated
 		r.diskData[newSnapName].Created = created
 		r.diskData[newSnapName].RevisionCounter = r.GetRevisionCounter()
+		// create new metafile for snapshot
 		if err := r.encodeToFile(r.diskData[newSnapName], newSnapName+metadataSuffix); err != nil {
 			return err
 		}
+		// crash here will leave stale snapshot files
```
payes, line 950: Put a TODO over here

replica/replica.go
```diff
@@ -901,34 +909,32 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
...
 
-	info := r.info
-	info.Head = newHeadDisk.Name
-	info.Dirty = true
-	info.Parent = newSnapName
-
-	if err := r.encodeToFile(&info, volumeMetaData); err != nil {
-		return err
-	}
+	// crash at this point leads to stale snapshot files
```
payes, line 936: Same TODO as below needs to be added here also

replica/replica.go
utkarshmani1997: yes, check GetReplica in REST Api calls

replica/replica.go
```diff
@@ -846,10 +846,7 @@ func (r *Replica) linkDisk(oldname, newname string) error {
 
 	dest := r.diskPath(newname)
 	if _, err := os.Stat(dest); err == nil {
```
vishnuitta, line 848: we are supposed to have similar check for meta file also

replica/replica.go
```diff
@@ -883,16 +880,15 @@ func (r *Replica) rmDisk(name string) error {
 		return nil
 	}
 
-	lastErr := os.Remove(r.diskPath(name))
-	if err := os.Remove(r.diskPath(name + metadataSuffix)); err != nil {
-		lastErr = err
+	if err := os.Remove(r.diskPath(name)); err != nil {
```
vishnuitta, line 883: we need to make notExists error as not error.. but, here its going to return error for 'notExists' also..
same comment applies for next one also..

utkarshmani1997:
@vishnuitta fixed those parts and added unit tests to verify the same.

PR #294, opened 2020-04-13
fix(replica): Fix issues with snapshot sync

payes:
Fixes: https://github.com/openebs/openebs/issues/2999

```
Problem Description:
According to current implementation:

R1          R2              R3
H1(s1)      H1(S1)          H1(S2)
H2(S2)      H2(S2)          H2 
H3          H3              

When R1, R2 and R3 are in sync, all of them have H1 as head file. Sequence of events that trigger rebuild time:

1. R3 gets disconnected
2. R2 gets disconnected
3. R2 comes back up again, therefore internal snapshot(S1) is created, and H1 is converted to S1 at R1 and R2. A new head(H2) is also created at R1 and R2.
4. R3 comes back up. A new snapshot S2 is created. H2 at R1 and R2 is converted to S2. And H1 at R3 is converted to S2.
5. Now data from S2 in R1 is copied to S2 in R3. And since there is no S1 in R3, S1 is also copied from R1 to R3.

Post fix scenario:
After this fix, problem reported in step 4 and 5 is fixed as described below.
When R3 comes up, a new snapshot is created. H2 at R1 and R2 is converted to S2. And H1 at R3 is converted to S1. Now data from S1 in R1 is synced with S1 in R3. And since there is no S2 in R3, S3 is copied from R1 to R3.

Proposed changes:
1. No snapshot name should be passed to replicas while creating auto snapshots.
- Previously a randomly generated name was passed.
After this change:
volume-head-000.img converts to volume-snap-000.img
volume-head-101.img converts to volume-snap-101.img
- This ensures that snapshot files at all the replicas have been created by renaming the same head file. This case was not always valid earlier as demonstrated above.

2. New head name should be passed while creating the snapshot. This head name is created by increasing the current head fetched from healthy replica by 1.
- This again ensures that all the head files being written to at the replicas are same. When converting this head file to snap, snapnames will be consistent at all the replicas.

After applying above 2 proposed changes:

R1          R2              R3
H1(s1)      H1(S1)          H1(S1)
H2(S2)      H2(S2)          
H3          H3              H3
	   
3. Sync head metadata file after syncing all the other snapshots at replica.
- Previously head metadata file was being synced by controller before starting the sync process in prepareRebuild function. That was not a valid approach, since if replica crashes after this step, chain should have been broken. But was not a problem since latest snapshot names were same at all the replicas. Head metadata file points to latest snapshot as that is its parent.

4. There was a bug in increasing the numbers in head file name which caused the numbers to circle after 10.
- After volume-head-010.img, the new head which was being created was volume-head-001.img instead of volume-head-011.img.

5. For manually created snapshots, snapshot name is passed by controller to replica but the image file name will still be corresponding to the head name as described above. But the name inside the corresponding metadata file will be same as snapshot name given by the user for revert, clone and snapshot delete purposes.

Upgrade Instructions:
Head metadata file is now being synced by replica, therefore we need to upgrade replicas before controller. If controller is upgraded before replica, head metadata file might not be synced before making replica to RW.

```

### DIFF
--- a/controller/rebuild.go
+++ b/controller/rebuild.go
@@ -24,6 +24,21 @@ func getReplicaChain(address string) ([]string, error) {
 	return rep.Chain, nil
 }
 
+func getReplicaDiskInfo(address string) (map[string]types.DiskInfo, error) {
+	repClient, err := client.NewReplicaClient(address)
+	if err != nil {
+		return nil, fmt.Errorf("Cannot get replica client for %v: %v",
+			address, err)
+	}
+
+	rep, err := repClient.GetReplica()
+	if err != nil {
+		return nil, fmt.Errorf("Cannot get replica for %v: %v",
+			address, err)
+	}
+	return rep.Disks, nil
+}
+
 func (c *Controller) getCurrentAndRWReplica(address string) (*types.Replica, *types.Replica, error) {
 	var (
 		current, rwReplica *types.Replica
--- a/controller/rebuild.go
+++ b/controller/rebuild.go
@@ -166,17 +166,5 @@ func (c *Controller) PrepareRebuildReplica(address string) ([]string, error) {
 		return nil, err
 	}
 
-	fromHead := rwChain[0]
-
-	chain, err := getReplicaChain(address)
-	if err != nil {
-		return nil, err
-	}
-	toHead := chain[0]
-
-	if err := syncFile(fromHead+".meta", toHead+".meta", rwReplica, replica); err != nil {
-		return nil, err
-	}
-
-	return rwChain[1:], nil
+	return rwChain, nil
 }
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -230,7 +230,7 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 	} else if size <= 0 {
 		return nil, os.ErrNotExist
 	} else {
-		if err := r.createDisk("000", false, util.Now()); err != nil {
+		if err := r.createDisk("", "volume-head-000.img", false, util.Now()); err != nil {
 			return nil, err
 		}
 	}
@@ -248,7 +248,7 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 }
 
 func GenerateSnapshotDiskName(name string) string {
-	return fmt.Sprintf(diskName, name)
+	return strings.Replace(name, headPrefix, diskPrefix, 1)
 }
 
 func GetSnapshotNameFromDiskName(diskName string) (string, error) {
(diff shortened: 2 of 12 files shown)


replica/replica.go
```diff
@@ -990,7 +998,9 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 		}
 
 		r.updateChildDisk(oldHead, newSnapName)
-		r.activeDiskData[len(r.activeDiskData)-1].Name = newSnapName
+		// Not Required as diskData and activeDiskData are pointing to same
+		// disks
+		//r.activeDiskData[len(r.activeDiskData)-1].Name = newSnapName
```
utkarshmani1997, line 1003: activeDiskData keeps the file in order, whereas diskData is a map which keeps the data in random order.

utkarshmani1997: This is required since it is used in snapshot deletion.

replica/rest/model.go
```diff
@@ -43,6 +43,7 @@ type RebuildingInput struct {
 type SnapshotInput struct {
 	client.Resource
 	Name        string `json:"name"`
```
utkarshmani1997, line 45: For creating the snapshot do we need to pass the headname as well or controller will take care of this?

controller/rebuild.go
```diff
@@ -166,17 +181,5 @@ func (c *Controller) PrepareRebuildReplica(address string) ([]string, error) {
...
-		return nil, err
-	}
-	toHead := chain[0]
-
-	if err := syncFile(fromHead+".meta", toHead+".meta", rwReplica, replica); err != nil {
-		return nil, err
-	}
-
-	return rwChain[1:], nil
+	return rwChain, nil
```
utkarshmani1997, line 184: We can skip prepare rebuild from WO replica, as we already have healthy replica info, so we can get the chain info from there, calling controller to perform rebuilding seems to be unnecessary as controller is not maintaining info.

sync/sync.go
```diff
@@ -278,8 +286,14 @@ Register:
 		return err
 	}
 
+	logrus.Infof("syncFiles from:%v to:%v", fromClient, toClient)
 	if !ok {
-		logrus.Infof("syncFiles from:%v to:%v", fromClient, toClient)
+		if err = t.syncFiles(fromClient, toClient, output.Disks); err != nil {
```
utkarshmani1997, line 291: What will happen if healthy replica will be of older version?

sync/sync.go
```diff
@@ -278,8 +286,14 @@ Register:
...
+			return err
+		}
+	} else {
+		// Only sync head metadata
+		output.Disks = output.Disks[:1]
```
utkarshmani1997, line 296: Why do we need to sync head meta when chain is same? 
isn't parent in both replicas be same?

replica/replica.go
```diff
@@ -975,7 +979,11 @@ func (r *Replica) createDisk(name string, userCreated bool, created string) erro
 	if newSnapName != "" {
 		r.addChildDisk(newSnapName, newHeadDisk.Name)
 		r.diskData[newSnapName] = r.diskData[oldHead]
-		r.diskData[newSnapName].Name = newSnapName
+		if userCreated {
+			r.diskData[newSnapName].Name = fmt.Sprintf(diskName, name)
```
utkarshmani1997, line 983: It's not necessary that user provide snapshot name while creation, could you verify `./jivactl snapshot create` commands ?

replica/rest/model.go
payes: Controller will take care of this

payes:
Closing this PR as it is not a priority at this point of time.

PR #296, opened 2020-04-21
fix(replica): make preload operations of secondary replicas lockless

payes:
This PR addresses the issue of preload taking time for secondary replicas causing write ios from controller to timeout. This issue was initially tried to solve by increasing the write timeout from 15 to 30s which is not scalable.
This PR solves the same issue as follows:
Preload will be done only once.
    - In case of 1st replica, preload is done when the replica gets a start signal from controller.
    - In case of remaining replicas, preload is done after completing the sync operation.

### DIFF
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -213,7 +213,7 @@ func drainHoleCreatorChan() {
 }
 
 //CreateHoles removes the offsets from corresponding sparse files
-func CreateHoles() error {
+func CreateHoles() {
 	var (
 		fd uintptr
 	)
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -327,6 +327,7 @@ func preload(d *diffDisk) error {
 
 func PreloadLunMap(d *diffDisk) error {
 	logrus.Info("Start reading extents")
+	inject.AddPreloadTimeout()
 	if err := preload(d); err != nil {
 		return err
 	}
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -293,6 +293,7 @@ func preload(d *diffDisk) error {
 				if d.files[d.location[offset]] != file ||
 					offset != lOffset+length {
 					if file != nil && fileIndx > userCreatedSnapIndx && shouldCreateHoles() {
+						d.UsedBlocks -= length
 						sendToCreateHole(file, lOffset*d.sectorSize, length*d.sectorSize)
 					}
 					file = d.files[d.location[offset]]
@@ -304,6 +305,8 @@ func preload(d *diffDisk) error {
 					// block will be punched outside the loop
 					length++
 				}
+			} else {
+				d.UsedLogicalBlocks++
 			}
 			d.location[offset] = uint16(i)
 			d.UsedBlocks++
@@ -311,6 +314,7 @@ func preload(d *diffDisk) error {
 		// This will take care of the case when the last call in the above loop
 		// enters else case
 		if file != nil && fileIndx > userCreatedSnapIndx && shouldCreateHoles() {
+			d.UsedBlocks -= length
 			sendToCreateHole(file, lOffset*d.sectorSize, length*d.sectorSize)
 		}
 		file = nil
@@ -330,12 +334,6 @@ func PreloadLunMap(d *diffDisk) error {
 	inject.AddPreloadTimeout()
 	if err := preload(d); err != nil {
 		return err
-	}
-
-	for _, val := range d.location {
-		if val != 0 {
-			d.UsedLogicalBlocks++
-		}
 	}
 	logrus.Info("Read extents successful")
 	return nil
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -280,12 +280,6 @@ func preload(d *diffDisk) error {
 		if i == 0 {
 			continue
 		}
-		if i == 1 {
-			// Reinitialize to zero so that we can detect holes in the base snapshot
-			for j := 0; j < len(d.location); j++ {
-				d.location[j] = 0
-			}
-		}
 		if d.UserCreatedSnap[i] {
 			userCreatedSnapIndx = uint16(i)
 		}
@@ -332,6 +326,7 @@ func preload(d *diffDisk) error {
 }
 
 func PreloadLunMap(d *diffDisk) error {
+	logrus.Info("Start reading extents")
 	if err := preload(d); err != nil {
 		return err
 	}
@@ -341,5 +336,6 @@ func PreloadLunMap(d *diffDisk) error {
 			d.UsedLogicalBlocks++
 		}
 	}
+	logrus.Info("Read extents successful")
 	return nil
 }
--- a/replica/backup.go
+++ b/replica/backup.go
@@ -286,6 +286,8 @@ func preload(d *diffDisk) error {
 		generator := newGenerator(d, f)
 		for offset := range generator.Generate() {
 			if d.location[offset] != 0 {
+				// d.UsedBlocks is being incremented and decremented to accomodate user
+				// created snapshots
 				// We are looking for continuous blocks over here.
 				// If the file of the next block is changed, we punch a hole
 				// for the previous unpunched blocks, and reset the file and
--- a/replica/client/client.go
+++ b/replica/client/client.go
@@ -125,6 +125,9 @@ func (c *ReplicaClient) Revert(name, created string) error {
 	if err != nil {
 		return err
 	}
+	// TODO Revisit this
+	// This is being done since Revert operation includes preload of lunMap at
+	// replica which might take time
 	c.httpClient.Timeout = 0
 	return c.post(r.Actions["revert"], rest.RevertInput{
 		Name:    name,
--- a/replica/client/client.go
+++ b/replica/client/client.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/openebs/jiva/replica/rest"
 	"github.com/openebs/jiva/sync/agent"
+	"github.com/openebs/jiva/util"
 	"github.com/sirupsen/logrus"
 )
 
@@ -128,6 +129,16 @@ func (c *ReplicaClient) Revert(name, created string) error {
 	return c.post(r.Actions["revert"], rest.RevertInput{
 		Name:    name,
 		Created: created,
+	}, nil)
+}
+
+func (c *ReplicaClient) SetLogging(lf util.LogToFile) error {
+	r, err := c.GetReplica()
+	if err != nil {
+		return err
+	}
+	return c.post(r.Actions["setlogging"], rest.LoggingInput{
+		LogToFile: lf,
 	}, nil)
 }
 
--- a/replica/client/client.go
+++ b/replica/client/client.go
@@ -124,7 +124,7 @@ func (c *ReplicaClient) Revert(name, created string) error {
 	if err != nil {
 		return err
 	}
-
+	c.httpClient.Timeout = 0
 	return c.post(r.Actions["revert"], rest.RevertInput{
 		Name:    name,
 		Created: created,
--- a/replica/diff_disk.go
+++ b/replica/diff_disk.go
@@ -182,6 +182,7 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 		offset = startSector + i
 		if val := d.location[startSector+i]; val == 0 {
 			d.UsedLogicalBlocks++
+			d.UsedBlocks++
 		} else if val != uint16(target) {
 			// d.UsedBlocks is being incremented and decremented to accomodate user
 			// created snapshots
@@ -204,8 +205,10 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 				//block will be punched outside the loop
 				length++
 			}
+			d.UsedBlocks++
 		}
-		d.UsedBlocks++
+		// Control will come over here if offsets are overwritten in the same
+		// file
 		d.location[startSector+i] = uint16(target)
 	}
 	//This will take care of the case when the last call in the above loop
--- a/replica/diff_disk.go
+++ b/replica/diff_disk.go
@@ -188,10 +188,11 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 			//If the file of the next block is changed, we punch a hole
 			//for the previous unpunched blocks, and reset the file and
 			//fileIndx pointed to by this block
+			// No need to decrease used blocks while punching holes as same
+			// number needs to be incremented
 			if d.location[startSector+i] != fileIndx ||
 				startSector+i != lOffset+length {
 				if file != nil && int(fileIndx) > d.SnapIndx && shouldCreateHoles() && !inject.DisablePunchHoles() {
-					d.UsedBlocks -= length
 					sendToCreateHole(d.files[val], lOffset*d.sectorSize, length*d.sectorSize)
 				}
 				file = d.files[d.location[startSector+i]]
@@ -209,7 +210,6 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
 	//This will take care of the case when the last call in the above loop
 	//enters else case
 	if (file != nil) && (int(fileIndx) > d.SnapIndx) && shouldCreateHoles() && !inject.DisablePunchHoles() {
-		d.UsedBlocks -= length
 		sendToCreateHole(file, lOffset*d.sectorSize, length*d.sectorSize)
 	}
 	file = nil
--- a/replica/diff_disk.go
+++ b/replica/diff_disk.go
@@ -16,7 +16,7 @@ type fileType struct {
 }
 
 type diffDisk struct {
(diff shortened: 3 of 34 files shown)


replica/backup.go
```diff
@@ -280,12 +280,6 @@ func preload(d *diffDisk) error {
 		if i == 0 {
 			continue
 		}
-		if i == 1 {
-			// Reinitialize to zero so that we can detect holes in the base snapshot
-			for j := 0; j < len(d.location); j++ {
```
utkarshmani1997, line 285: Why this has been removed?

replica/client/client.go
```diff
@@ -124,7 +124,7 @@ func (c *ReplicaClient) Revert(name, created string) error {
 	if err != nil {
 		return err
 	}
-
+	c.httpClient.Timeout = 0
```
utkarshmani1997, line 127: Setting timeout to 0 means this request may hung forever, till the connection is available.

replica/replica.go
```diff
@@ -238,12 +238,11 @@ func construct(readonly bool, size, sectorSize int64, dir, head string, backingF
 
 	r.insertBackingFile()
 	r.ReplicaType = replicaType
-	logrus.Info("Start reading extents")
-	inject.AddPreloadTimeout()
-	if err := PreloadLunMap(&r.volume); err != nil {
-		return r, fmt.Errorf("failed to load Lun map, error: %v", err)
+	if types.PreloadDuringOpen {
```
utkarshmani1997, line 241: lets donot use global variables, it introduces potential bugs hard to catch.

replica/rest/model.go
```diff
@@ -173,8 +173,6 @@ func NewReplica(context *api.ApiContext, state replica.State, info replica.Info,
 		actions["updatecloneinfo"] = true
 	case replica.Rebuilding:
 		actions["start"] = true
-		actions["resize"] = true
-		actions["snapshot"] = true
```
utkarshmani1997, line 177: let's remove resize and snapshot from other invalid states also, or put a TODO about the same

sync/sync.go
```diff
@@ -171,10 +171,17 @@ func (t *Task) CloneReplica(url string, address string, cloneIP string, snapName
 		if err != nil {
 			return fmt.Errorf("Failed to update clone info, err: %v", err)
 		}
+		types.PreloadDuringOpen = false
```
utkarshmani1997, line 174: same comment as earlier.

replica/backup.go
payes: This is not required as the map is always empty whenever preload is called.

replica/client/client.go
utkarshmani1997: Why this change in revert?

payes: If the connection is available, we expect the server to respond. This is a revert operation and all the replicas should respond with success or failure before further steps could be taken.

utkarshmani1997:
@payes we have set timeout to 0 in Open call from controller as well, we need to fix it by other ways like retrying by increasing timeout in other iterations instead of erroring out in first go. This is safe and best that we can do IMO.

replica/replica.go
payes: This variable will be removed in the near future when we remove preload() from open() function.
Multiple function signatures will have to be changed to embed this variable in replica structure. But if you think it might create bugs in short term, I can make this change ?

replica/client/client.go
payes: Revert also does a preload operation which may time out. Therefore this change was necessary.

replica/replica.go
vishnuitta: better to pass as argument or in a structure

payes:
@vishnuitta These have been disabled in NewReplica() func in rebuilding state.

replica/server.go
```diff
@@ -360,6 +360,74 @@ func (s *Server) Revert(name, created string) error {
...
+				holeLength = 1
+				holeOffset = offset
+				prevHoleFileIndx = fileIndx
+			} else {
+				holeLength++
+			}
+		} else {
+			// No hole drilling over here as that offset is empty
+			s.r.volume.location[offset] = volume.location[offset]
+			if prevHoleFileIndx > userCreatedSnapIndx && shouldCreateHoles() && prevHoleFileIndx != 0 {
```
vishnuitta, line 420: do we have a testcase where we hit this?

replica/server.go
```diff
@@ -360,6 +360,74 @@ func (s *Server) Revert(name, created string) error {
...
+				}
+				holeLength = 1
+				holeOffset = offset
+				prevHoleFileIndx = fileIndx
+			} else {
+				holeLength++
+			}
+		} else {
+			// No hole drilling over here as that offset is empty
+			s.r.volume.location[offset] = volume.location[offset]
```
vishnuitta, line 419: we need to be copying other variables of diffDisk like UsedBlocks and UsedLogicalBlocks and testcases for them

replica/server.go
vishnuitta: this is a good one

replica/server.go
vishnuitta: Also, it seems some problem in calculating UsedBlocks of old code.. we are NOT decrementing it when the holes are punched

replica/backup.go
vishnuitta: lets add comments accordingly @payes with the description, assumptions of inputs and the output of this function

replica/server.go
```diff
@@ -145,7 +154,7 @@ func (s *Server) Create(size int64) error {
 	sectorSize := s.getSectorSize()
 
 	logrus.Infof("Creating volume %s, size %d/%d", s.dir, size, sectorSize)
-	r, err := New(size, sectorSize, s.dir, s.backing, s.ServerType)
+	r, err := New(false, size, sectorSize, s.dir, s.backing, s.ServerType)
```
vishnuitta, line 157: a comment here would help on why we don't need preload

replica/server.go
```diff
@@ -54,6 +55,14 @@ func NewServer(dir string, sectorSize int64, serverType string) *Server {
 	}
 }
 
+// SetPreload sets/unsets preloadDuringOpen flag
+func (s *Server) SetPreload(preload bool) error {
+	s.Lock()
+	defer s.Unlock()
+	s.preload = true
```
vishnuitta, line 62: this is always setting the preload.. it needs fix

replica/server.go
```diff
@@ -39,6 +39,7 @@ type Server struct {
 	//the replica attempts to connect back to controller
 	MonitorChannel chan struct{}
 	//closeSync      chan struct{}
+	preload bool
```
vishnuitta, line 42: instead of a 'server' variable, can we consider by parameter passing only through all functions? this also becomes a kind of global variable

vishnuitta: if we are considering to go with this member variable, we need to set this as 'true' by default, and whenever this is unset, it need to be reset to true to assume that this is 'true' by default
However, I prefer passing as argument due to reason mentioned above

payes: preload is sent by open() call from controller. If we want to pass it via function arguments again we need to make changes to REST API calls.

sync/sync.go
```diff
@@ -171,10 +171,17 @@ func (t *Task) CloneReplica(url string, address string, cloneIP string, snapName
...
 		if err != nil {
 			return fmt.Errorf("Failed to reload clone replica, error: %s", err.Error())
 		}
+		// Preload was not needed to be done separately in case of clone as IOs
+		// are not going on but in case if clone functionality is enhanced to
+		// multiple cloned replicas, then this might be needed
+		if err := s.UpdateLUNMap(); err != nil {
+			return fmt.Errorf("UpdateLUNMap() failed, err: %v", err.Error())
+		}
 		if err := toClient.SetRebuilding(false); err != nil {
```
vishnuitta, line 185: here it can be reset to 'true'

payes: done

replica/server.go
payes: fixed

sync/sync.go
```diff
@@ -171,10 +171,18 @@ func (t *Task) CloneReplica(url string, address string, cloneIP string, snapName
 		if err != nil {
 			return fmt.Errorf("Failed to update clone info, err: %v", err)
 		}
+		s.SetPreload(false)
 		_, err = toClient.ReloadReplica()
 		if err != nil {
 			return fmt.Errorf("Failed to reload clone replica, error: %s", err.Error())
 		}
```
vishnuitta, line 178: we can set it to true here itself

sync/sync.go
```diff
@@ -239,6 +247,7 @@ Register:
 		return fmt.Errorf("CheckAndResetFailedRebuild failed, error: %s", err.Error())
 	}
 
+	s.SetPreload(false)
 	logrus.Infof("Adding replica %s in WO mode", replicaAddress)
 	_, err = t.client.CreateReplica(replicaAddress)
 	if err != nil {
```
vishnuitta, line 253: we can reset it to true and add a comment saying 'after operations like create, reload, preload will be reset to its default value i.e., true.. same is followed at other places where these operations are performed'

sync/sync.go
```diff
@@ -336,12 +345,16 @@ func (t *Task) checkAndResetFailedRebuild(address string, server *replica.Server
 	return nil
 }
 
-func (t *Task) reloadAndVerify(address string, repClient *replicaClient.ReplicaClient) error {
+func (t *Task) reloadAndVerify(s *replica.Server, address string, repClient *replicaClient.ReplicaClient) error {
 	_, err := repClient.ReloadReplica()
```
vishnuitta, line 349: set it to fase before and true later

sync/sync.go
```diff
@@ -263,6 +263,7 @@ Register:
 		time.Sleep(5 * time.Second)
 		return err
 	}
+	s.SetPreload(true)
```
vishnuitta, line 266: set this to above the 'if' condition.. there is a return from there

replica/rest/model.go
```diff
@@ -180,9 +181,6 @@ func NewReplica(context *api.ApiContext, state replica.State, info replica.Info,
 		actions["setreplicacounter"] = true
 		actions["updatecloneinfo"] = true
 	case replica.Rebuilding:
-		actions["start"] = true
```
vishnuitta, line 183: can we remove many other actions from Rebuilding state? (or) do we need to allow them so that if replica restart happens after the state is set?

payes: these actions are required during rebuild

sync/sync.go
payes: done

sync/sync.go
payes: done

replica/replica_test.go
```diff
@@ -1004,3 +1007,84 @@ func (s *TestSuite) TestPartialReadZeroEndOffset(c *C) {
 
 	byteEquals(c, expected, buf)
 }
+
```
vishnuitta, line 1010: some comments to understand what testcase it is about, even before getting into code to get context

replica/server.go
```diff
@@ -443,6 +445,7 @@ func (s *Server) UpdateLUNMap() error {
 				sendToCreateHole(volume.files[prevHoleFileIndx], int64(holeOffset)*volume.sectorSize, holeLength*volume.sectorSize)
 			}
 			holeOffset = 0
+			holeLength = 0
```
vishnuitta, line 448: good find

sync/sync.go
```diff
@@ -196,6 +203,10 @@ func (t *Task) AddReplica(replicaAddress string, s *replica.Server) error {
 	if s == nil {
 		return fmt.Errorf("Server not present for %v, Add replica using CLI not supported", replicaAddress)
 	}
+	logrus.Infof("CheckAndResetFailedRebuild %v", replicaAddress)
```
vishnuitta, line 206: do we have a testcase for this change? existing testcases may cover this as well if I'm not wrong

vishnuitta: this is another good one

vishnuitta: I thought this is good one as 'rebuilding' status will be set to false, but, the function consists of a condition along with 'state'.. and, there seems to be a Open call inside this.. this is going to take time..
so, why we need this change?

vishnuitta: I mean - it should be because we removed 'start' action when rebuild is set.. but, there is condition along with 'state' to be closed.. will that be the case always?
And, having 'Open' call in this is going to impact time taken for first replica to become connected and RW?

sync/sync.go
```diff
@@ -234,13 +254,10 @@ Register:
 		}
 		return nil
 	}
-	logrus.Infof("CheckAndResetFailedRebuild %v", replicaAddress)
-	if err := t.checkAndResetFailedRebuild(replicaAddress, s); err != nil {
-		return fmt.Errorf("CheckAndResetFailedRebuild failed, error: %s", err.Error())
-	}
-
 	logrus.Infof("Adding replica %s in WO mode", replicaAddress)
+	s.SetPreload(false)
```
vishnuitta, line 258: some comment here by saying 'CreateReplica goes to controller and it triggers Open call with replica.. but we don't want to perform preload in this case.. so setting Preload to false

replica/server.go
```diff
@@ -360,6 +372,93 @@ func (s *Server) Revert(name, created string) error {
 	return nil
 }
 
+// UpdateLUNMap updates the the original LUNmap with any changes which have been
```
vishnuitta, line 375: hopefully this function can be called multiple times

sync/sync.go
payes: done

replica/server.go
payes: added comment

sync/sync.go
payes: Added setting preload to false in this. And also removed the extra condition which will always be true.

replica/replica_test.go
payes: added comment

replica/diff_disk.go
```diff
@@ -182,15 +182,17 @@ func (d *diffDisk) fullWriteAt(buf []byte, offset int64) (int, error) {
...
+			// created snapshots
+			// We are looking for continuous blocks over here.
+			// If the file of the next block is changed or offset not same,
+			// we punch a hole for the previous unpunched blocks,
+			// and reset the file and fileIndx pointed to by this block
 			if d.location[startSector+i] != fileIndx ||
 				startSector+i != lOffset+length {
-				if file != nil && int(fileIndx) > d.SnapIndx && shouldCreateHoles() && !inject.DisablePunchHoles() {
+				if (file != nil) && (int(fileIndx) > d.SnapIndx) && shouldCreateHoles() && !inject.DisablePunchHoles() {
+					d.UsedBlocks -= length
```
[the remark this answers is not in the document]
payes: done

replica/server.go
```diff
@@ -360,6 +372,93 @@ func (s *Server) Revert(name, created string) error {
...
+				extraLogicalBlocks++
+			}
+			continue
+		}
+		if s.r.volume.location[offset] > fileIndx {
+			// It is being incremented and decremented to accomodate user
+			// created snapshots
+			if prevHoleFileIndx != fileIndx || int64(offset) != int64(holeOffset)+holeLength {
+				if prevHoleFileIndx > userCreatedSnapIndx && shouldCreateHoles() && prevHoleFileIndx != 0 {
+					extraUsedBlocks -= holeLength
```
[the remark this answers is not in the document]
vishnuitta: why to increment Logical blocks? Its not the new blocks.. its overwrites

replica/server.go
```diff
@@ -360,6 +372,93 @@ func (s *Server) Revert(name, created string) error {
...
+					sendToCreateHole(volume.files[prevHoleFileIndx], int64(holeOffset)*volume.sectorSize, holeLength*volume.sectorSize)
+				}
```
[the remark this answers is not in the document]
payes: These are logical offsets, they need to be converted to actual file offsets to punch holes.

replica/replica_test.go
```diff
@@ -1004,3 +1007,84 @@ func (s *TestSuite) TestPartialReadZeroEndOffset(c *C) {
...
+func (s *TestSuite) TestUpdateLUNMap(c *C) {
```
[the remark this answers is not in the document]
payes: I will add them in upcoming PRs.

replica/server.go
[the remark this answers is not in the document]
utkarshmani1997: What i understand from the code is UsedBlocks are the blocks used by all the files whereas logical blocks are where actual data is written. So both needs to be updated or both should not be updated. @payes can put more light on it.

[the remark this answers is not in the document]
payes: If a particular offset is filled in the `volume`, UsedLogical block count need to be incremented.
If a particular offset is filled in a `file`, Used block count need to be incremented.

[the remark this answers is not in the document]
utkarshmani1997: Thanks @payes got it.

PR #301, opened 2020-05-07
fix(sync): make sync http client timeout configurable

utkarshmani1997:
Add env SYNC_HTTP_CLIENT_TIMEOUT to make sync client http request timeout configurable.
This env can be passed as container argument.


utkarshmani1997:
Right

PR #308, opened 2020-05-21
replica(leader-election): Avoid making replica in rebuilding state as master

payes:


payes:
This PR is not needed anymore as the fix for sync has been made in https://github.com/openebs/jiva/pull/316

PR #311, opened 2020-05-24
refactor(build): fix travis errors on forked repo

kmova:
When building on Travis with IMAGE_REPO env set
to forked repo, the CI scripts were failing
due to a hardcoding of container registry org name as openebs.

This commit helps to re-use the IMAGE ORG
specified in the TRAVIS ENV.

If ENV is not specified, defaults to `openebs`. 

### DIFF
--- a/ci/suite.sh
+++ b/ci/suite.sh
@@ -1,4 +1,3 @@
-
 #!/bin/bash
 set -x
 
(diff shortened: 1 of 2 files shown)


kmova:
Tested with ENV variable set on a forked repo here: https://github.com/mayadata-io/jiva/pull/1

PR #313, opened 2020-05-25
refactor(build): make the org configurables

kmova:
Removing a few more hardcoded references to openebs, seeing during pushing jiva docker image. 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,11 @@ IS_DOCKER_INSTALLED       := $(shell which docker >> /dev/null 2>&1; echo $$?)
 
 TARGETS := $(shell ls scripts)
 
+ifeq (${IMAGE_ORG}, )
+  IMAGE_ORG = openebs
+  export IMAGE_ORG
+endif
+
 # Determine the arch/os
 ifeq (${XC_OS}, )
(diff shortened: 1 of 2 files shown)


kmova:
Tested via: https://github.com/mayadata-io/jiva/pull/2

Makefile
```diff
@@ -13,6 +13,11 @@ IS_DOCKER_INSTALLED       := $(shell which docker >> /dev/null 2>&1; echo $$?)
 
 TARGETS := $(shell ls scripts)
 
+ifeq (${IMAGE_ORG}, )
+  IMAGE_ORG = openebs
```
vishnuitta, line 17: is this supposed to be ":="? seeing this at all the other places

kmova: Checked this on. #TIL. 
When making direct string assignments, `:=` is not required. not required.  `:=`  is used in other places to expand expression during definition.

vishnuitta: both the forms does the expansion.. '=' makes recursive ones which can lead to infinite loop, where as, ':=' does expansion only once..
for direct assignments, both should be fine (https://www.gnu.org/software/make/manual/make.html#Flavors)

PR #316, opened 2020-05-31
feat(auto-delete-snapshot): add auto-snapshot deletion below checkpoint

payes:
This PR re-introduces the auto-snapshot deletion feature for auto-generated snapshots.

Approach:
1. Mark a snapshot as committed. Over here committed means all the replicas are synced till this point. We don’t need to sync snapshots created before this.
2. Snapshot is marked as committed by the controller once the replication factor number of replicas are in RW.
3. Now we can delete snapshots below this point without any issue.
4. This reduces our sync time also.
5. To perform deletion of snapshot, replica checks with controller to make sure that they are all having same checkpointed snapshot.


vishnuitta:
also mention in description that: `To perform deletion of snapshot, replica checks with other replicas to make sure that they are all having same checkpointed snapshot`

controller/control.go
```diff
@@ -134,6 +134,16 @@ func (c *Controller) UpdateVolStatus() {
...
+			if err := c.backend.SetCheckpoint(checkpoint); err == nil {
+				c.Checkpoint = checkpoint
+			}
+		}
+	} else {
+		c.Checkpoint = ""
+	}
 
 	logrus.Infof("Previously Volume RO: %v, Currently: %v,  Total Replicas: %v,  RW replicas: %v, Total backends: %v",
 		prev, c.ReadOnly, len(c.replicas), rwReplicaCount, len(c.backend.backends))
```
vishnuitta, line 149: print the checkpoint here.. previous checkpoint also

controller/replicator.go
```diff
@@ -312,6 +312,94 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+				} else {
+					retErrorLock.Lock()
+					replicaChains[addr] = chain
+					success = success + 1
+					retErrorLock.Unlock()
+				}
+				wg.Done()
+			}(addr, backend.backend)
+		} else {
+			logrus.Infof("not taking snapshot at %s in err mode", addr)
```
vishnuitta, line 342: this 'else' case as well should be treated as error

controller/replicator.go
```diff
@@ -312,6 +312,94 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+		}
+	}
+	wg.Wait()
+
+	if len(retError.Errors) != 0 {
+		return "", retError
+	}
+	var snapName string
+	for _, chain := range replicaChains {
```
vishnuitta, line 351: should we make sure that chains are also same at all the places?
and also should check for replicationFactor also matching with len(replicaChains)?

controller/control.go
```diff
@@ -134,6 +134,16 @@ func (c *Controller) UpdateVolStatus() {
 	} else {
 		c.ReadOnly = true
 	}
+	if rwReplicaCount == c.ReplicationFactor {
+		checkpoint, err := c.backend.GetLastSnapshot()
+		if err == nil {
```
vishnuitta, line 139: what if 'err' != nil? need to reset Checkpoint as it is got reset if rwReplicaCount is not equal to c.ReplicationFacotr?

controller/control.go
```diff
@@ -134,6 +134,16 @@ func (c *Controller) UpdateVolStatus() {
...
+			if err := c.backend.SetCheckpoint(checkpoint); err == nil {
```
vishnuitta, line 140: this 'err' as well need to be handled?

controller/replicator.go
```diff
@@ -312,6 +312,94 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+					retErrorLock.Unlock()
+				} else {
+					retErrorLock.Lock()
+					success = success + 1
+					retErrorLock.Unlock()
+				}
+				wg.Done()
+			}(addr, backend.backend)
+		} else {
+			logrus.Infof("not setting checkpoint at %s in err mode", addr)
```
vishnuitta, line 391: here as well need to be considered as error

controller/control.go
```diff
@@ -134,6 +134,16 @@ func (c *Controller) UpdateVolStatus() {
 	} else {
 		c.ReadOnly = true
 	}
+	if rwReplicaCount == c.ReplicationFactor {
```
vishnuitta, line 137: are we doing this with the lock?

replica/rest/model.go
```diff
@@ -194,6 +202,7 @@ func NewReplica(context *api.ApiContext, state replica.State, info replica.Info,
 		actions["setrevisioncounter"] = true
 		actions["setreplicacounter"] = true
 		actions["updatecloneinfo"] = true
+		actions["setcheckpoint"] = true
```
vishnuitta, line 205: I don't think we need this in Rebuilding state.. we might also not needed in Dirty state..
Is Open the state of replica after it becomes RW? only that state need to be allowed

replica/replica.go
```diff
@@ -1413,3 +1401,12 @@ func (r *Replica) SetReplicaMode(mode string) error {
 	}
 	return nil
 }
+
+// SetCheckpoint ...
+func (r *Replica) SetCheckpoint(snapshotName string) error {
+	r.Lock()
+	defer r.Unlock()
+	r.info.Checkpoint = snapshotName
```
vishnuitta, line 1409: don't we need to check that snapshotName is part of the chain?

controller/control.go
```diff
@@ -1207,25 +1217,17 @@ func (c *Controller) IsReplicaRW(replicaInController *types.Replica) error {
 }
 
 // DeleteSnapshot ...
-func (c *Controller) DeleteSnapshot(replicas []types.Replica) error {
+func (c *Controller) DeleteSnapshot(snapshot string, replicas []types.Replica) error {
```
vishnuitta, line 1220: is DeleteSnapshot triggered by controller? its supposed to be by replica right

sync/sync.go
```diff
@@ -289,52 +299,66 @@ Register:
...
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
 	logrus.Infof("RevisionCount of RW replica (%v): %v, Current replica (%v): %v",
 		fromClient.GetAddress(), rwReplica.RevisionCounter, toClient.GetAddress(),
 		curReplica.RevisionCounter)
 	logrus.Infof("RW replica chain: %v, cur replica chain: %v", rwReplica.Chain, curReplica.Chain)
+
+	if curReplica.Checkpoint != "" {
```
vishnuitta, line 338: should we make sure checkpoints are also same?

payes: Here, while syncing it is not required and not necessary to be same. Suppose while creating checkpoint the replica crashed, both replicas will not have same checkpoint in that case.
What can be done over here is, data can be synced after the checkpoint of the RW Replica, since if RW Replica has a checkpoint that snapshot has to be present in WO replica also.
But for a replica from scratch, i.e, if a replica is not having checkpoint, whole data should be synced.

controller/control.go
payes: This function is not being used, will be removed. From controller snapshot deletion can be triggered manually via cli.

replica/replica.go
payes: This check is performed at the controller in GetLastSnapshot(). But it can be checked at replica also.

replica/rest/model.go
payes: As soon as an IO is written after opening the state becomes dirty.
It has been allowed in rebuilding state because this will be triggered as a part of verifyRebuildReplica for WO replica. While calling verifyRebuildReplica, replica is in Rebuilding state.

controller/control.go
payes: Yes

controller/replicator.go
payes: will be adding a check to match Replication factor with replicaChains.
Matching the chains is already performed when a replica is converted to RW mode in verifyRebuildReplica().

payes:
This PR has been split into the following PRs: https://github.com/openebs/jiva/pull/319 and https://github.com/openebs/jiva/pull/321

PR #318, opened 2020-06-08
fix(pattern): fix pattern to avoid repetition of head file names after every 10 snapshots

payes:
Head file names were getting repeated after every 10 snapshots. Although this was not causing any  problems as of now but fixing this will help in future implementations where these head file names will be used to sync data. Besides the repetition was not intended.

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -36,7 +36,7 @@ const (
 )
 
 var (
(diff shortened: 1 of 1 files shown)


vishnuitta:
or can this change be postponed till the need arises?

payes:
@vishnuitta Even though this doesn't cause any problem, it also creates an inconsistent view of the meta files while debugging issues.

PR #319, opened 2020-06-13
feat(checkpoint): add checkpoint functionality and optimized rebuild

payes:
Approach:
1. Mark a snapshot as committed. Over here committed means all the replicas are synced till this point. We don’t need to sync snapshots created before this.
2. Snapshot is marked as committed by the controller once the replication factor number of replicas are in RW.
3. Rebuild is done only for the snapshots created after this checkpoint


controller/control.go
```diff
@@ -135,8 +137,18 @@ func (c *Controller) UpdateVolStatus() {
 		c.ReadOnly = true
 	}
 
-	logrus.Infof("Previously Volume RO: %v, Currently: %v,  Total Replicas: %v,  RW replicas: %v, Total backends: %v",
-		prev, c.ReadOnly, len(c.replicas), rwReplicaCount, len(c.backend.backends))
+	if rwReplicaCount == c.ReplicationFactor {
+		lastSnapshot, err := c.backend.GetLastSnapshot()
+		if err == nil {
+			if err := c.backend.SetCheckpoint(checkpoint); err == nil {
```
vishnuitta, line 143: here, should we be setting `lastSnapshot` as checkpoint right?

controller/control.go
```diff
@@ -135,8 +137,18 @@ func (c *Controller) UpdateVolStatus() {
 		c.ReadOnly = true
 	}
 
-	logrus.Infof("Previously Volume RO: %v, Currently: %v,  Total Replicas: %v,  RW replicas: %v, Total backends: %v",
-		prev, c.ReadOnly, len(c.replicas), rwReplicaCount, len(c.backend.backends))
+	if rwReplicaCount == c.ReplicationFactor {
```
vishnuitta, line 140: lets use this function to just UpdateVolStatus.
Make a new function for this code, and call it in the caller with probably any conditions required.
You can have new conditions in new function also.

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
 	return nil
 }
 
+func (r *replicator) GetLastSnapshot() (string, error) {
+	retErrorLock := sync.Mutex{}
+	retError := &BackendError{
+		Errors: map[string]error{},
```
vishnuitta, line 318: don't we need to perform make of this 'map'?

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+	replicaChains := make(map[string][]string)
+	wg := sync.WaitGroup{}
+	var success int
+
+	for addr, backend := range r.backends {
+		if backend.mode == types.RW {
+			wg.Add(1)
+			go func(address string, backend types.Backend) {
+				if chain, err := backend.GetReplicaChain(); err != nil {
+					logrus.Infof("failed taking snapshot at %s with err %v", addr, err)
```
vishnuitta, line 329: change the error message

controller/replicator.go
vishnuitta: make sure we hit error case in testing

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
 	return nil
 }
 
+func (r *replicator) GetLastSnapshot() (string, error) {
+	retErrorLock := sync.Mutex{}
+	retError := &BackendError{
```
vishnuitta, line 317: why a struct over map? can't we just create a map `retErrors map[string][]string` similar to replicaChains

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+			go func(address string, backend types.Backend) {
+				if chain, err := backend.GetReplicaChain(); err != nil {
+					logrus.Infof("failed taking snapshot at %s with err %v", addr, err)
+					retErrorLock.Lock()
+					retError.Errors[address] = err
+					retErrorLock.Unlock()
+				} else {
+					retErrorLock.Lock()
+					replicaChains[addr] = chain
+					success = success + 1
```
vishnuitta, line 336: we don't seems to be using this

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
 	return nil
 }
 
+func (r *replicator) GetLastSnapshot() (string, error) {
```
vishnuitta, line 315: is this means Latest snapshot? Please provide comments

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+	wg.Wait()
+	if len(retError.Errors) != 0 {
+		return "", retError
+	}
+	if len(replicaChains) != len(r.backends) {
+		return "", fmt.Errorf("Not able to fetch chain from all replicas")
+	}
+	var snapName string
+	for _, chain := range replicaChains {
+		if len(chain) <= 1 {
```
vishnuitta, line 354: at one replica being <=1 and at other replicas being > 1 is possible. Instead of 'continue' doing 'break' is better?

vishnuitta: and by setting snapName to ""

vishnuitta: or return error also if caller can't handle empty snapName and nil error

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+					retErrorLock.Unlock()
+				} else {
+					retErrorLock.Lock()
+					success = success + 1
+					retErrorLock.Unlock()
+				}
+				wg.Done()
+			}(addr, backend.backend)
+		} else {
+			logrus.Infof("not setting checkpoint at %s in err mode", addr)
```
vishnuitta, line 393: this as well should be treated as error

controller/replicator.go
```diff
@@ -312,6 +312,95 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+				} else {
+					retErrorLock.Lock()
+					replicaChains[addr] = chain
+					success = success + 1
+					retErrorLock.Unlock()
+				}
+				wg.Done()
+			}(addr, backend.backend)
+		} else {
+			logrus.Infof("not getting chain from %s as replica in %v mode", addr, backend.mode)
```
vishnuitta, line 342: this as well a error and break

vishnuitta: testcase should cover this

controller/control.go
```diff
@@ -135,8 +137,18 @@ func (c *Controller) UpdateVolStatus() {
...
+		lastSnapshot, err := c.backend.GetLastSnapshot()
```
vishnuitta, line 141: it is latestSnapshot

payes: updating the name to latest snaphsot

controller/replicator.go
payes: This is handled below in:
```if len(replicaChains) != len(r.backends) {```
adding break laos over here

File: controller/replicator.go (hunk 8 above)
Comment (payes) [reply controller/replicator.go:354]: updating to return error. This was done primarily to handle single replica case. But yes this was a bug.

File: controller/replicator.go (hunk 6 above)
Comment (payes) [reply controller/replicator.go:336]: removed

File: controller/replicator.go (hunk 5 above)
Comment (payes) [reply controller/replicator.go:317]: This has to be returned as an error. That  is why it has been been in a struct. It is a common practice being followed in other functions also.

File: controller/replicator.go (hunk 3 above)
Comment (payes) [reply controller/replicator.go:318]: This is another way to initialize a map. [initialize map in go](https://stackoverflow.com/questions/31064688/which-is-the-nicer-way-to-initialize-a-map-in-golang)

File: controller/control.go (hunk 1 above)
Comment (payes) [reply controller/control.go:143]: Fixed it. Typo while splitting PRs.

File: controller/control.go (hunk 2 above)
Comment (payes) [reply controller/control.go:152]: done

File: controller/control.go [hunk 12]
```diff
@@ -134,9 +135,33 @@ func (c *Controller) UpdateVolStatus() {
...
+
+	if rwReplicaCount == c.ReplicationFactor {
+		latestSnapshot, err := c.backend.GetLatestSnapshot()
+		if err == nil {
+			if err := c.backend.SetCheckpoint(latestSnapshot); err == nil {
+				checkpoint = latestSnapshot
+			}
+		}
+	}
+	c.Checkpoint = checkpoint
```
vishnuitta, line 162: if 'err' happens in any of the above statement, c.Checkpoint gets set to "".. is that the expected?
can you put the comments for the function with i/p, o/p and the functionality?

controller/control.go
```diff
@@ -134,9 +135,33 @@ func (c *Controller) UpdateVolStatus() {
...
+
+	logrus.Infof("prevCheckpoint: %v, currCheckpoint: %v", prevCheckpoint, c.Checkpoint)
```
vishnuitta, line 164: printing err would be good if any

replica/replica.go
```diff
@@ -1413,3 +1414,12 @@ func (r *Replica) SetReplicaMode(mode string) error {
...
 	return nil
 }
+
+// SetCheckpoint ...
+func (r *Replica) SetCheckpoint(snapshotName string) error {
+	r.Lock()
+	defer r.Unlock()
+	r.info.Checkpoint = snapshotName
+
+	return r.encodeToFile(&r.info, volumeMetaData)
```
vishnuitta, line 1424: hope we have a testcase for upgrade scenario

replica/server.go
```diff
@@ -645,6 +646,19 @@ func (s *Server) SetReplicaMode(mode string) error {
...
 }
 
+// SetCheckpoint ...
+func (s *Server) SetCheckpoint(snapshotName string) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if s.r == nil {
+		logrus.Infof("s.r is nil during setCheckpoint")
+		return nil
```
vishnuitta, line 656: returning 'err' here is better if I'm not wrong

controller/control.go
payes: Yes that is the expectation. Adding comment for the same.

replica/replica.go
payes: Upgrade test is already present, still adding one more test to upgrade from 1.11 release

vishnuitta:
@utkarshmani1997 would you be able to review these changes?

controller/control.go
```diff
@@ -134,9 +135,46 @@ func (c *Controller) UpdateVolStatus() {
...
+	logrus.Infof("Previously Volume RO: %v, Currently: %v, Total Replicas: %v, RW replicas: %v, Total backends: %v",
+		prevState, c.ReadOnly, len(c.replicas), rwReplicaCount, len(c.backend.backends))
+}
+
+// UpdateCheckpoint updates c.Checkpoint based on the last snapshot.
+// This function should be called after every removal/addition of replicas.
+// GetLatestSnapshot and SetCheckpoint will error out in cases when one of
+// the replicas go down in which case we should unset c.Checkpoint
+// Checkpoint should be set only when
+// replication factor number of replicas are in RW state
```
vishnuitta, line 149: thanks.. this comment helps.
Did we considered the case of adding a replica or removing a replica and the manual steps for that?

payes: While replacing the replicas, the checkpoint for new replica is nil, in which case all the snapshots are synced. On removing the old replica, checkpoint at controller is automatically set to nil.
Manual steps are required when auto deletion is triggered which will be pushed in auto-deletion PR.

sync/sync.go
```diff
@@ -311,30 +311,44 @@ Register:
...
 	curReplica, err := toClient.GetReplica()
 	if err != nil {
-		return false, err
+		return false, nil, err
 	}
 	logrus.Infof("RevisionCount of RW replica (%v): %v, Current replica (%v): %v",
 		fromClient.GetAddress(), rwReplica.RevisionCounter, toClient.GetAddress(),
 		curReplica.RevisionCounter)
 	logrus.Infof("RW replica chain: %v, cur replica chain: %v", rwReplica.Chain, curReplica.Chain)
+	if curReplica.Checkpoint != "" {
```
vishnuitta, line 328: lets add the comments here why there is no need to check for below the checkpoint?

vishnuitta: if I'm not wrong, replica should confirm on checkpoint before proceeding for deletion?

payes: Yes replica verifies the checkpoint from controller before proceeding for snapshot deletion. This will be pushed in proceeding PR.

controller/rebuild.go
```diff
@@ -76,17 +91,29 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
+		snapshot string
+	)
+	for indx, snapshot = range rwChain {
+		if snapshot == WOcheckpoint {
+			break
+		}
+	}
+	logrus.Infof("chain %v from rw replica %s", rwChain, rwReplica.Address)
+	// No need to compare the volume head disk
+	rwChain = rwChain[1 : indx+1]
```
vishnuitta, line 113: WOCheckPoint can be empty also, and so, indx + 1 can lead to out of boundary

sync/sync.go
```diff
@@ -311,30 +311,44 @@ Register:
...
+		for indx, snapshot := range curReplica.Chain {
+			if snapshot == curReplica.Checkpoint {
+				curReplica.Chain = curReplica.Chain[:indx]
```
vishnuitta, line 335: should the Checkpoint related one also need to be compared?

replica/server.go
```diff
@@ -381,6 +386,9 @@ func (s *Server) UpdateLUNMap() error {
 	// be having 2 LUNMaps, one being filled by the parallel write operations
 	// and the other being filled by preload operation.
 	s.Lock()
+	if s.r == nil {
+		return fmt.Errorf("UpdateLUNMap failed, s.r not set")
```
vishnuitta, line 390: unlock required before returning error

sync/sync.go
payes: @vishnuitta Not able to grasp this comment

controller/rebuild.go
payes: in rwChain[1:indx+1], indx+1 element is not included. It is similar to [1:indx+1).

payes: https://play.golang.org/p/JhEmlx8FHMb

utkarshmani1997:
@vishnuitta i'm occupied with some of the things right now. I will review the changes over this weekend 😊

vishnuitta:
> @vishnuitta i'm occupied with some of the things right now (It's little overwhelming as i'm learning new things). I will review the changes over this weekend blush
> @payes could you please share the design documents by mentioning what are changes are going in with this PR as it looks big (43 files have been changed).

sounds good @utkarshmani1997 .. thanks

payes:
@utkarshmani1997 Design doc is mentioned in description [Snapshot deletion in jiva](https://docs.google.com/document/d/1Ka_vG34ZxzEA5qCr8utrFxCNjIS7ZCW2H7fPjZPqZHY/edit?usp=sharing)

controller/rebuild.go
```diff
@@ -76,17 +91,29 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
+	var (
+		indx     int
+		snapshot string
+	)
+	for indx, snapshot = range rwChain {
+		if snapshot == WOcheckpoint {
+			break
+		}
+	}
+	logrus.Infof("chain %v from rw replica %s", rwChain, rwReplica.Address)
```
vishnuitta, line 111: lets print index as well (and in next print also)

controller/rebuild.go
```diff
@@ -76,17 +91,29 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
+	WOcheckpoint, err := getReplicaCheckpoint(address)
+	if err != nil {
+		return err
+	}
+	var (
+		indx     int
+		snapshot string
+	)
+	for indx, snapshot = range rwChain {
+		if snapshot == WOcheckpoint {
```
vishnuitta, line 107: Should we check that WOcheckpoint should exist in rwChain if WOcheckpoint is not empty?

controller/rebuild.go
```diff
@@ -76,17 +91,29 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
 
-	logrus.Infof("chain %v from rw replica %s", rwChain, rwReplica.Address)
-	// Don't need to compare the volume head disk
-	rwChain = rwChain[1:]
-
 	chain, err := getReplicaChain(address)
 	if err != nil {
 		return err
 	}
+	WOcheckpoint, err := getReplicaCheckpoint(address)
```
vishnuitta, line 98: can you add a comment on why WOcheckpoint should be considered but not that of RWcheckpoint?

vishnuitta: consider possible cases in the comment like WO checkpoint is latest than RW checkpoint and vice versa

sync/sync.go
```diff
@@ -311,30 +311,48 @@ Register:
...
+				break
+			}
+		}
+		for indx, snapshot := range rwReplica.Chain {
+			if snapshot == curReplica.Checkpoint {
+				rwReplica.Chain = rwReplica.Chain[:indx]
```
vishnuitta, line 341: print 'idx' also

sync/sync.go
```diff
@@ -311,30 +311,48 @@ Register:
...
+	// having same data. Checkpoints are created only when the controller
+	// is certain that till this snapshot all the replicas are in sync.
+	if curReplica.Checkpoint != "" {
+		for indx, snapshot := range curReplica.Chain {
+			if snapshot == curReplica.Checkpoint {
+				curReplica.Chain = curReplica.Chain[:indx]
+				break
+			}
+		}
+		for indx, snapshot := range rwReplica.Chain {
```
vishnuitta, line 339: a check is required that curReplica Checkpoint should exists in rwReplica Chain.. else throw error
Similar one required in VerifyRebuildReplica

controller/replicator.go
```diff
@@ -312,6 +312,98 @@ func (r *replicator) Snapshot(name string, userCreated bool, created string) err
...
+		}
+		if snapName == "" {
+			snapName = chain[1]
+			continue
+		}
+		if snapName != chain[1] {
+			return "", fmt.Errorf("Latest Snapshot differs at replicas %v", replicaChains)
+		}
+	}
+	return snapName, nil
```
utkarshmani1997, line 364: Though checkpoints are same across replicas, it is possible that chain is not same.
Need to check if chain is matching across replicas before returning the check pointed snapshot.
Please clarify if i'm missing something and put a comment why this is required.

controller/control.go
```diff
@@ -295,14 +333,8 @@ func (c *Controller) addQuorumReplica(address string, snapshot bool) error {
 		return fmt.Errorf("Failed to set rebuild : %v", true)
 	}
 	c.UpdateVolStatus()
+	c.UpdateCheckpoint()
```
utkarshmani1997, line 336: I think we should remove Quorum related code.

controller/control.go
```diff
@@ -689,6 +724,7 @@ func (c *Controller) RemoveReplicaNoLock(address string) error {
 		}
 	}
 	c.UpdateVolStatus()
+	c.UpdateCheckpoint()
```
utkarshmani1997, line 727: Why do we need to update the checkpoint while removing replicas?

payes: Checkpoint is maintained at the controller to facilitate snapshot deletion at the replicas.
Before deleting an auto-generated snapshot, replica checks with the controller if checkpoint is set.
If checkpoint is set at the controller, it indicates that all the replicas are in sync, and this checkpoint is set at all the replicas.
Now any replicas can delete a snapshot below this checkpoint.
When a replica needs to be replaced, snapshot deletion process is the RW replicas need to be stopped. The only way to do it is by unsetting the checkpoint at the controller as soon as a replica goes down.

controller/control.go
payes: agreed, this is a long pending task which need to be completed.

controller/replicator.go
payes: This function is called when all the replicas are in rw, in which case the latest snapsohts in all the replicas should be same. And chain above the checkpoint will also be same, since all are in sync. Chain below the checkpoint might be different, but we are not concerned about that.

sync/sync.go
payes: Checkpointed snapshot need not be compared as it should be present at all the replicas

controller/rebuild.go
```diff
@@ -95,24 +96,27 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
 	var (
 		indx     int
 		snapshot string
 	)
+	if (WOCheckpoint != "") && (!util.ChainContainsSnapshot(rwChain, WOCheckpoint)) {
+		return fmt.Errorf("WO replica's checkpoint not present in rwReplica chain")
+	}
 	for indx, snapshot = range rwChain {
-		if snapshot == WOcheckpoint {
+		if snapshot == WOCheckpoint {
```
vishnuitta, line 111: does index '0' of rwChain hae snapshot as ""? does index 0 needs special handling?

controller/rebuild.go
```diff
@@ -95,24 +96,27 @@ func (c *Controller) VerifyRebuildReplica(address string) error {
...
-	WOcheckpoint, err := getReplicaCheckpoint(address)
+	WOCheckpoint, err := getReplicaCheckpoint(address)
 	if err != nil {
 		return err
 	}
 	var (
 		indx     int
 		snapshot string
 	)
+	if (WOCheckpoint != "") && (!util.ChainContainsSnapshot(rwChain, WOCheckpoint)) {
```
vishnuitta, line 107: comments are better @payes ..

sync/sync.go
```diff
@@ -325,10 +325,23 @@ func (t *Task) isRevisionCountAndChainSame(fromClient, toClient *replicaClient.R
 		fromClient.GetAddress(), rwReplica.RevisionCounter, toClient.GetAddress(),
 		curReplica.RevisionCounter)
 	logrus.Infof("RW replica chain: %v, cur replica chain: %v", rwReplica.Chain, curReplica.Chain)
```
vishnuitta, line 327: lets print checkpoints also here

controller/rebuild.go
vishnuitta: ok.. got to see them in sync/sync.go

sync/sync.go
payes: printed checkpoint in if condition, which will be printed if checkpoint is not nil

controller/rebuild.go
payes: index 0 is head file

payes: Index 0 doesn't need to be compared, as head file name might be different across replicas

PR #321, opened 2020-07-19
feat(auto-delete-snapshot): add auto-snapshot deletion below checkpoint

payes:
This PR re-introduces the auto-snapshot deletion feature for auto-generated snapshots.

Approach:
1. A deletion thread keeps on running in the background at each replica and executes after every 60s
2. It fetches the checkpoint from controller which is present at in-memory of controller.
3. If checkpoint is not available at controller, replica will not initiate a delete.
4. Replica creates a deletion candidate list which contains all the snapshots created before the checkpointed snapshot.
5. This list is sorted based on the size.
6. In each iteration of the thread, one snapshot is picked from the top and deleted.

Replacing a replica:
1. Remove the desired replica.
2. Set checkpoint to nil at controller. This will already have become nil as soon as the replica was disconnected.
3. This REST Call inturn verifies that snapshot deletion is not under progress at any of the connected replicas.
4. Add the new replica.

Notes:
1. Snapshot Deletion request at controller will only mark the snapshot as removed at the replicas. These snapshots will be deleted in the background by internal Snapshot deletion thread.
2. Snapshot list will show all snapshots instead of just the common snapshots. Snapshots marked as removed will be skipped from this list.


sync/sync.go
```diff
@@ -421,6 +431,7 @@ func (t *Task) reloadAndVerify(s *replica.Server, address string, repClient *rep
 		return err
 	}
 
+	go t.InternalSnapshotCleaner(s, repClient)
```
vishnuitta, line 434: lets move this to end of AddReplica rather than keeping here

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+// It fetches the checkpoint from controller which is present at in-memory of controller.
+// If checkpoint is not available at controller, snapshot delete is not initiated.
+// A deletion candidate list is prepared
+// In each iteration of the loop, one snapshot is picked from the top and deleted.
+func (t *Task) InternalSnapshotCleaner(s *replica.Server, repClient *replicaClient.ReplicaClient) {
+	ticker := time.NewTicker(SnapshotDeletionInterval)
+
+	for range ticker.C {
+		snapshot, err := t.client.GetCheckpoint()
+		if err == nil && snapshot == s.Replica().Info().Checkpoint {
```
vishnuitta, line 642: perform continue if err is not nil or snapshots are not equal
and how to handle if snapshots are not equal?

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+		replicaChain = append(replicaChain, chain[i])
+	}
+	replicaDisks := s.Replica().ListDisks()
+
+	for indx, snapshot = range replicaChain {
+		if snapshot == checkpoint {
+			break
+		}
+	}
+	if len(replicaChain) <= 3 || indx <= 1 { // Head, last snapshot, base snapshot
```
vishnuitta, line 724: you can move this check on length to above after getting the replica Chain. And, indx can be very high and still can be returned NULL i.e., when checkpoint is empty.
So, special handling of empty checkpoint is also needed.

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+		snapshot string
+	)
+
+	chain, err := s.Replica().Chain()
+	if err != nil {
+		return nil, err
+	}
+	var replicaChain []string
+	for i := len(chain) - 1; i >= 0; i-- {
+		replicaChain = append(replicaChain, chain[i])
```
vishnuitta, line 715: some comments here how replicaChain looks like.. i.e., from base to head

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+	}
+	replicaDisks := s.Replica().ListDisks()
+
+	for indx, snapshot = range replicaChain {
+		if snapshot == checkpoint {
```
vishnuitta, line 720: handle case where this check never succeeds

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+}
+
+// getDeleteCandidateChain returns the chain of snapshots that can be deleted
+// Chain is sorted based on the snapshot size before returning
+// All the snapshots in the chain are returned except:
+// 1. Head snapshot
+// 2. Last snapshot, snapshot just below head
+// 3. Base snapshot
+// 4. User created snapshots not marked as removed
+func getDeleteCandidateChain(s *replica.Server, checkpoint string) ([]SnapList, error) {
```
vishnuitta, line 702: can we have unit tests just for this function? cases like
- empty checkpoint
- chain length of 3
- not existing checkpoint
- user created and not removed
- user created and removed
- varying size of snapshots to get the sorted one

vishnuitta: I guess - hitting these cases helps, and can give us few more scenarios.

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+		}
+		snapList[i].name = disk
+		snapList[i].size, err = strconv.ParseInt(replicaDisks[disk].Size, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to convert size: %v into int64, err: %v", replicaDisks[disk].Size, err)
+		}
+		i++
+	}
+
+	sort.SliceStable(snapList, func(i, j int) bool {
```
vishnuitta, line 743: what happens if nothing to added to snapList?

vishnuitta: I mean - should we handle it by returning nil, nil, the way above checks are there for len.
(or) we can remove both len related checks as caller is taking care of them.

sync/sync.go
```diff
@@ -617,3 +628,121 @@ func (t *Task) getToReplica(address string) (rest.Replica, error) {
...
+	for range ticker.C {
+		snapshot, err := t.client.GetCheckpoint()
+		if err == nil && snapshot == s.Replica().Info().Checkpoint {
+			sortedSnapshotList, _ := getDeleteCandidateChain(s, snapshot)
+			if len(sortedSnapshotList) < 10 {
+				continue
+			}
+			var snap SnapList
+			for _, snap = range sortedSnapshotList {
+				if snap.name != "" {
```
vishnuitta, line 649: can snap.name possible to be empty?

replica/replica.go
```diff
@@ -563,9 +563,12 @@ func (r *Replica) PrepareRemoveDisk(name string) ([]PrepareRemoveAction, error)
 		disk = GenerateSnapshotDiskName(name)
 		data, exists = r.diskData[disk]
 		if !exists {
-			return nil, fmt.Errorf("Can not find snapshot %v", disk)
+			return nil, nil
 		}
 	}
+	if data.Removed {
```
vishnuitta, line 569: so, if the disk is already in removed state, then, it never gets deleted.
case like where disk is marked as removed as part of deletion process, and replica got disconnected.

replica/replica.go
```diff
@@ -563,9 +563,12 @@ func (r *Replica) PrepareRemoveDisk(name string) ([]PrepareRemoveAction, error)
 		disk = GenerateSnapshotDiskName(name)
 		data, exists = r.diskData[disk]
 		if !exists {
-			return nil, fmt.Errorf("Can not find snapshot %v", disk)
+			return nil, nil
```
vishnuitta, line 566: how this is not error case? and moreover, why snapshot disk is generated here?

replica/replica.go
```diff
@@ -606,37 +612,19 @@ func (r *Replica) processPrepareRemoveDisks(disks []string) ([]PrepareRemoveActi
...
-		children := r.diskChildrenMap[disk]
-		// 1) leaf node
-		if children == nil {
-			actions = append(actions, PrepareRemoveAction{
+		parent := r.diskData[disk].Parent
+
+		actions = append(actions,
+			PrepareRemoveAction{
+				Action: OpCoalesce,
+				Source: disk,
```
vishnuitta, line 620: is this way of coalescing supported by the 'sync' binary?
I remember that there is a reason for copying from a disk to its child, rather than copying from a disk to its parent.

replica/replica.go
```diff
@@ -606,37 +612,19 @@ func (r *Replica) processPrepareRemoveDisks(disks []string) ([]PrepareRemoveActi
...
+				Target: parent,
+			},
+			PrepareRemoveAction{
 				Action: OpRemove,
```
vishnuitta, line 624: there are checks for count of the children in removeDiskNode to error out if children count is > 1.
would that fine or should be handled?

replica/replica.go
vishnuitta, line 620: if there is swap between source and target of the output, then, the existing callers of PrepareRemoveDisk need to be modified as well.

controller/control.go
```diff
@@ -1244,25 +1244,17 @@ func (c *Controller) IsReplicaRW(replicaInController *types.Replica) error {
...
 	for _, r := range replicas {
 		replica := r // pin it
-		ops[replica.Address], err = c.prepareRemoveSnapshot(&replica, c.SnapshotName)
+		_, err = c.prepareRemoveSnapshot(&replica, snapshot)
 		if err != nil {
 			return err
 		}
 	}
 
-	for _, rep := range replicas {
```
vishnuitta, line 1259: don't understand why this is removed

vishnuitta: as per PR message, deleting snapshots is left to snapshot cleanup go routine.
But, PrepareRemoveDisk is returning nil, nil when removed is true.

payes: When a snapshot delete request is received, prepareRemoveDisk is being called which marks the disk as removed.
Once this disk is marked as removed, this is taken up by InternalSnapshotDeletion thread to delete the snapshot.

replica/diff_disk.go
```diff
@@ -62,7 +62,7 @@ func (d *diffDisk) RemoveIndex(index int) error {
 	}
 
 	for i := 0; i < len(d.location); i++ {
-		if d.location[i] > uint16(index) {
+		if d.location[i] >= uint16(index) {
```
vishnuitta, line 65: I don't think that this is correct. Assuming indexes are from head to base for 0 to n and 'i'th index need to be removed.
i'th data is copied to i+1, i.e., i'th parent. And, now, it becomes i'th one. It consists of old i'th's file content and also 'i+1''s content. So, i'th index will not be change.

replica/replica.go
payes: The output of prepare remove disk was only being called at controller to delete snapshot. Code at controller has been removed to Coalesce and Delete the snapshot.

payes: We are always copying data from child to parent, since child has the latest data.

replica/diff_disk.go
payes: Indexes are from base to head, head index is the largest, base index is 0

replica/replica.go
payes: If the disk is not present in the list, implies it has already been deleted. This check was valid when controller was triggering the delete. A snapshot might have been deleted from one replica but present in others. When replica deleted the snapshot by itself, this should never be hit.
GenerateSnapshotDiskName() just adds suffix and prefix to the snapshot name to get valid disk names.

replica/replica.go
payes: removed this check, thanks.

replica/diff_disk.go
vishnuitta: ok.. can you fix the comments according to new logic and add few more comments

vishnuitta: can you verify that it leads to corruption without this change and testcase catches it?

app/snapshot.go
```diff
@@ -234,6 +234,45 @@ func getCommonSnapshots(replicas []rest.Replica) ([]string, error) {
...
+			}
+		}
+
+		if !first {
+			diffSnapshots = util.Filter(snapshots, func(i string) bool {
+				return !util.Contains(finalSnapshotList, i)
+			})
+		}
+		first = false
+		finalSnapshotList = append(snapshots, diffSnapshots...)
```
vishnuitta, line 271: here, we might be trying to find common snapshots across all the replicas when all of them are in 'RW' state.
so, we should be using finalSnapshotList to append as well.

vishnuitta: ok.. we are getting all the snapshots, not just the common ones.
Why do we need them? and if so, its better to return the replica addresses also where it is currently available?

vishnuitta: but, we should still use finalSnapshotList to append and make some other changes.

sync/sync.go
payes: It cannot be empty over here. This has been added just as a precausion

sync/sync.go
payes: Empty checkpoint case never happens as it has been taken care in the caller (InternalSnapshotCleaner()) when getting checkpoint from controller. But for function completion I will add here also.

sync/sync.go
payes: added panic after 3 continuous mismatches

sync/sync.go
payes: empty snapList passed to sort.SliceStable() doesn't cause any problem. 
If snapList is empty, empty list is returned which should not be a problem.

app/snapshot.go
payes: Will add replica address in a separate PR, fixed the above issue.

sync/sync.go
payes: added these tests

replica/replica.go
payes: handled

replica/diff_disk.go
payes: can I add this is next PRs

replica/replica.go
```diff
@@ -237,6 +237,9 @@ func construct(preload, readonly bool, size, sectorSize int64, dir, head string,
 	}
 	r.info.Parent = r.diskData[r.info.Head].Parent
 
+	if err = r.removeStaleFromChildrenMap(); err != nil {
```
vishnuitta, line 240: why do we want to add and then to remove? can't we avoid adding in first place?

sync/sync.go
```diff
@@ -660,20 +664,14 @@ func (t *Task) InternalSnapshotCleaner(s *replica.Server, repClient *replicaClie
...
 		}
-		var snap SnapList
-		for _, snap = range sortedSnapshotList {
-			if snap.name != "" {
-				break
-			}
-		}
-		if snap.name == "" {
+		if sortedSnapshotList[0] == "" {
 			continue
```
vishnuitta, line 672: a log here would help, though this should never happen

replica/replica.go
payes: readMetadata() func adds them to list, till that point chain is not formed. Chain is formed in openLiveChain() which is called later.

replica/replica.go
```diff
@@ -1158,15 +1168,23 @@ func (r *Replica) readMetadata() (bool, error) {
 			} else {
 				r.volume.UsedBlocks += (file.Size()/defaultSectorSize + 1)
 			}
+			if r.info.Head == "" {
+				return false, fmt.Errorf("r.info.Head is nil")
+			}
+			file = fileMap[r.info.Head+metadataSuffix]
 		} else if strings.HasSuffix(file.Name(), metadataSuffix) {
-			if err := r.readDiskData(file.Name()); err != nil {
+			if parent, err = r.readDiskData(file.Name()); err != nil {
```
vishnuitta, line 1176: err is not handled

vishnuitta: fine.. its handled

PR #328, opened 2020-09-20
chore(license): add license-check for .go , .sh , Dockerfile and Makefile

payes:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,20 @@
+# Copyright © 2020 The OpenEBS Authors
+#
+# This file was originally authored by Rancher Labs
+# under Apache License 2018.
+#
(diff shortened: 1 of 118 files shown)


Makefile
```diff
@@ -1,3 +1,20 @@
+# Copyright © 2020 The OpenEBS Authors
+#
+# This file was originally authored by Rancher Labs
+# under Apache License 2018.
```
prateekpandey14, line 4: We can do something like this ?
```
Copyright © 2020 The Rancher Labs and OpenEBS Authors
```

kmova (approved):

Makefile
kmova: The above message looks fine Prateek. As the Rancher Labs authors are currently not in the maintainers list for jiva.

PR #332, opened 2020-11-11
refact(builds): enable amd64 and arm64 multi arch builds

prateekpandey14:
Signed-off-by: prateekpandey14 <prateekpandey14@gmail.com>

**Why is this PR required? What issue does it fix?**:
Enabling more than two arch in the multi-arch builds is resulting in increased build times as well as
intermittent failures in running the builds.

Limiting to two archs - amd64 and arm64 for now.


kmova:
@prateekpandey14 - can you also remove the jobs for arm and ppc builds from travis. https://github.com/openebs/jiva/blob/master/.travis.yml#L34

kmova (approved):

PR #337, opened 2020-12-12
fix(metaFiles): open MetaFiles with O_SYNC flag

payes:
This PR along with https://github.com/openebs/sparse-tools/pull/7 fixes the following issue: https://github.com/openebs/jiva/issues/338
On crashing the kernel simultaneously at all the nodes, one of the replica fails to come up and enters into CrashLoopBackOff state with this error:
 ```failed to read disk data, error while unmarshalling file: volume-snap-1db04a94-0aa0-4404-bbbb-3641392d587f.img.meta```

This issue occured because metadata files are not synced to disk immediately. It is not seen with data files as those are opened with O_DIRECT flag.

As a fix, Metadata files are now being opened with O_SYNC flag.

### DIFF
--- a/replica/replica.go
+++ b/replica/replica.go
@@ -443,7 +443,7 @@ func (r *Replica) hardlinkDisk(target, source string) error {
 	if err := os.Link(r.diskPath(source), r.diskPath(target)); err != nil {
 		return fmt.Errorf("Fail to link %s to %s", source, target)
 	}
-	return r.syncDir()
+	return r.SyncDir()
 }
 
 // ReplaceDisk replace the source with target snapshot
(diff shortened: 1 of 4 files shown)


utkarshmani1997 (approved):
Thanks @payes LGTM

utkarshmani1997:
@payes was thinking about wrt data files as well, should we run sync before creating snapshots as well? 
That way we can be sure that data is flushed to file completely.
cc: @pawanpraka1

replica/replica.go
```diff
@@ -929,6 +929,9 @@ func (r *Replica) revertDisk(parent, created string) (*Replica, error) {
 }
 
 func (r *Replica) createDisk(name string, userCreated bool, created string) error {
+	if err := r.SyncDir(); err != nil {
```
pawanpraka1, line 932: why we are calling this in the beginning of the function?

payes: this function is called to create snapshots. To ensure that the snapshot file has all the data and metadata synced to disk we are syncing the directory before creating snapshot. When snapshots are created IOs are paused, therefore at this point there are no IOs running in parallel.

pawanpraka1 (approved):
looks good.

kmova (approved):

PR #343, opened 2021-01-26
fix(registration): add UUID field to avoid duplicate registration attempts

payes:
fixes: https://github.com/openebs/openebs/issues/3323
This PR implements the following:
1. Add UUID to each replica to detect if the same replica tries to register with a different IP address.
2. If a selected replica leader is not reachable, a new leader is selected.

Test performed: Delays were added at replicas while registration to reproduce the above issue and the fix was tested against the same.


utkarshmani1997 (commented):
The code looks good, but how will you handle upgrades?

kmova:
@payes - could you update the PR with the tests conducted on this change. Like @utkarshmani1997 mentioned, how is the upgrade handled - if replicas are upgraded prior to controller, will there be any issues?

payes:
Thanks, @utkarshmani1997 @kmova for reviewing the code.
The cloned replica also goes through the same process of adding to the controller, through the AddReplica() function.
Half upgraded setup also doesn't cause any problem. But still, I am adding a check for nil UUID.

utkarshmani1997 (approved):

kmova (approved):

PR #344, opened 2021-02-03
refact(UUID): include replica address while creating UUID hash

payes:
This PR implements the following:
1. Update go version to 1.14.7
2. Avoid using `Dir` global variable in replica package to aid unit testing.
3.  Use replica IP to create UUID hash

### DIFF
--- a/replica/server.go
+++ b/replica/server.go
@@ -20,6 +20,7 @@
 package replica
 
 import (
+	"crypto/sha1"
 	"errors"
 	"fmt"
 	"os"
@@ -27,6 +28,7 @@ import (
 	"time"
 
 	fibmap "github.com/frostschutz/go-fibmap"
+	guuid "github.com/google/uuid"
 	inject "github.com/openebs/jiva/error-inject"
 	"github.com/openebs/jiva/types"
 	"github.com/sirupsen/logrus"
@@ -44,7 +46,7 @@ const (
 type State string
 
 var ActionChannel chan string
-var Dir string
+
 var StartTime time.Time
 
 type Server struct {
@@ -52,7 +54,7 @@ type Server struct {
 	r                 *Replica
 	ReplicaAddress    string
 	ServerType        string
-	dir               string
+	Dir               string
 	defaultSectorSize int64
 	backing           *BackingFile
 	//This channel is used to montitor the IO connection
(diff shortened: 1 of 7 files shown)


kmova:
@payes - can you update the PR with details like why this change is required and what has been done?

replica/server.go
```diff
@@ -159,25 +160,45 @@ func (s *Server) isExtentSupported() error {
...
+	return nil
+}
+
 func (s *Server) Create(size int64) error {
 	s.Lock()
 	defer s.Unlock()
 	if err := s.isExtentSupported(); err != nil {
 		return err
 	}
+	defer s.initUUID()
```
kmova, line 189: Why is this in `defer`? Shouldn;t it be executed prior to unlocking?

replica/server.go
```diff
@@ -159,25 +160,45 @@ func (s *Server) isExtentSupported() error {
...
+	}
+	if r.info, err = ReadInfo(s.Dir); err != nil {
+		return err
+	}
+	if r.info.UUID == "" {
+		UUID := guuid.New().String() + s.ReplicaAddress
+		hash := sha1.New()
+		hash.Write([]byte(UUID))
+		r.info.UUID = fmt.Sprintf("%x", hash.Sum(nil))
+		r.writeVolumeMetaData(r.info.Dirty, r.info.Rebuilding)
```
kmova, line 178: While upgrade/rebuilding, will the information in the master replica be set to Dirty and Rebuilding?

replica/server.go
payes: Since the Unlock is also in differ, this will be executed before unlock (https://play.golang.org/p/-oUdvDmPhnT)

replica/server.go
payes: This function will be fetching whatever is already written on disk, updating UUID and rewriting the same.
Controller and replicas won't know that they are being upgraded so the states do not change during upgrade.
During rebuilding, master replica state is dirty and state of replica being rebuilt is set to rebuilding.

kmova: cool! #TIL

replica/server.go
kmova: cool! #TIL

kmova (approved):

